PR #3, opened 2015-05-29
English translation for documentation and example

rpspringuel:


jperon:
I really don't know how to thank you… Translating to (correct) English takes me such a time that I'm quickly discouraged when I have to !

rpspringuel:
Since English is my native language, I find it fairly easy to translate from some other language I know (French or Latin) into English than the other way around.  This is especially true when I don't have to construct a literal translation, but rather can use what sounds more natural to my ear.

Going in the other direction, however, is an absolute bear, so I can sympathize.  I just don't have the same ear for my non-native languages that allows me to tell intuitively that a passage is saying what I want it to say.

PR #9, opened 2015-12-24
Updated translation of README

rpspringuel:
I noticed that the README had been updated and so updated the translation.


jperon:
Thank you very much ! Late is better than never, as we say in France… I couldn't merge it before, as I was making a try in a monastery !

PR #14, opened 2017-09-18
Fix spacing in ly environment

mmeyn:
this removes an extra space created after the score in a ly environment


jperon:
Many thanks !

PR #31, opened 2018-01-17
Protrusion

### LINKED ISSUE #23: Add option to define left-margin
With a defined `left-margin` option in the lilypond-score it will be possible to explicitly align the *first barline* to the text margin, which is something I'd greatly prefer.

Any value set here should be added to line-width. So a left margin of 0.5cm should add 0.5cm to the line width.

uliska:
Closes #23 
I think this is ready for review now.

There are further things to consider from a user's point of view. With this PR protrusion is "the" behaviour of the package, but there should be other options, for example when the score does *not* fill the whole line width (e.g. centered, short music examples). But I think this could be discussed *after* merging this.

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -190,9 +190,9 @@ function write_tex(output, staffsize)
         end
     end
     local texoutput, nbre = content:gsub([[\includegraphics{]],
-        PROTRUSION..[[\includegraphics{]]..dirname(output))
+        [[\noindent]]..' '..PROTRUSION..[[\includegraphics{]]..dirname(output))
     PROTRUSION = ''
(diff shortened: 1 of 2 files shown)


jperon:
Very nice, thank you !

uliska:
It was a pleasure, especially from the point on where I wasn't faced with purely confusing errors upon each code change ;-)

I'm glad the idea of protrusion worked more or less straightforwardly - and especially reliably. But the issue of aligning scores into text has to be discussed more thoroughly, aesthetically, regarding user interface and regarding implementation. 
I will prepare an issue with some examples and invite "the public" to join the discussion - if you don't have any objections.

Issue #30, opened 2018-01-17
Handling LilyPond compilation failures


uliska:
When a LilyPond compilation fails a Lua stack trace occurs in `write_tex` because it tries to read non-existent files.

This should be handled by at least suppressing the error. Of course the user should *somehow* be notified. In `musicexamples` missing files don't cause an error but an information box is printed instead. I think something similar to that would be a good idea.

uliska:
> I prefer that an error be raised when a non-existent file is trying to be read. 

OK, *I* prefer not interrupting the compilation - so this clearly speaks for making *both* behaviours possible ;-)

> Whether the error is a lua stack trace or a more readable package error doesn't bother me much.

It definitely should be more readable, as that stack trace is pretty confusing for regular users.

> However, now that I think about it, I can also see the benefit of doing something like what `musicexamples` does for the purpose of creating placeholders in a skeletal document, 

That touches a point I've noticed these days. Every now and then I think about implementing something that `musicexamples` does (or plans to do). However, I think we should try to avoid duplication of concerns as much as possible, but this issue shows that it may be more complex than would be desirable. Basically `lyluatex` should *not* produce that yellow placeholder but do something that can make `musicexamples` do so. BUT: the point in keeping the two packages separate is to enable them to work without the other. So what would `lyluatex` do if `musicexamples` is *not* present?

Actually I think we should *not* think this through from the perspective of this current issue item but keep it in mind for a later discussion about the interaction of the two packages in general. I would like to get more familiar with `lyluatex` first, maybe making some more contributions, and then go through `musicexamples` again, probably updating it significantly (probably quite a lot of the code was still pretty naive ...). Then we should discuss the relation of the two packages more generally.

jperon:
Would you like to implement other things on this issue, or may we close it ?

uliska:
Ich will think about IT, I Had one Mode Idea.

uliska:
Actually it's not necessary to keep this one open.
@rpspringuel wants an error to occur (which is the default behaviour), and alternatively a box is printed (which is what I would usually want). 

If anybody ever wants some other behaviour they can open a new issue.

PR #38, opened 2018-01-21
WIP: #30: Correctly identify and handle failed LilyPond compilations

uliska:
This is a preliminary implementation of #30 that properly detects and handles cases where LilyPond fails to compile a score.

I *think* I correctly identify the case with both fragment and full-page scores (which is completely different), but the consequence is preliminary: at the current state of opening that pull request a failed score inserts a placeholder comment *in the output document*, and we should discuss what else we could do (so this isn't meant to be merged yet).

One suggestion that might be acceptable to both @rpspringuel 's and my perspective would be to rewrite the text that is written to LaTeX to include a file with a file name "compilation-failure.pdf" or similar. That way "plain" LaTeX would simply choke about the missing file while `musicexamples` would print that yellow box wtih "missing example file “compilation-failure” in it.

```
Add function is_compiled

This is first used to determine whether a score has to be recompiled.
Look for either OUTPUT.pdf (for fullpage) or a non-empty
OUTPUT-systems.tex (note that also failed LilyPond compilations
produce a -systems.tex file)

the -systems.tex file also has to be removed, otherwise next time it would
be considered as "already compiled".
```

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -59,16 +59,30 @@ function hash_output_filename(ly_code, line_width, staffsize)
     return TMP..'/'..filename
 end
 
+
+function is_compiled(output)
+    if output:find('fullpage') then
+        if lfs.isfile(output..'.pdf') then
+            return true else return false end
+    end
+
+    f = io.open(output..'-systems.tex')
+    if not f then return false end
+
+    head = f:read("*line")
+    if head == "% eof" then return false else return true end
+end
+
+
 function lilypond_fragment(ly_code, line_width, staffsize)
     line_width, staffsize = extract_size_arguments(line_width, staffsize)
     ly_code = ly_code:gsub('\\par ', '\n'):gsub('\\([^%s]*) %-([^%s])', '\\%1-%2')
     local output = hash_output_filename(ly_code, line_width, staffsize)
-    local compiled = false
-    if not lfs.isfile(output..'-systems.tex') then
+    local new_score = not is_compiled(output)
+    if new_score then
         compile_lilypond_fragment(ly_code, staffsize, line_width, output, include)
-        compiled = true
     end
-    write_tex(output, compiled)
+    write_tex(output, new_score)
 end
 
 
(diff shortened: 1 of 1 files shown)


jperon:
I think that, instead of `print` (l. 205), it would be good to use `err`. You have several examples of that in [this file](https://github.com/gregorio-project/gregorio/blob/master/tex/gregoriotex.lua) from `gregoriotex`. This function raises an error that stops compilation and lets the user know what to do.

uliska:
OK, I think the error message is very helpful with that. However, I get a stack trace in addition:

```
/usr/share/texliv
e/texmf-dist/tex/latex/base/ltluatex.lua:109: 
Module lyluatex Error: 
(lyluatex)             Score failed to compile, please check something.
(lyluatex)             on input line 12

stack traceback:
	[C]: in function 'error'
	/usr/share/texlive/texmf-dist/tex/latex/base/ltluatex.lua:109: in function </u
sr/share/texlive/texmf-dist/tex/latex/base/ltluatex.lua:108>
	(...tail calls...)
	...e/uliska/texmf/tex/latex/latex-git/lyluatex/lyluatex.lua:205: in function '
write_tex'
	...e/uliska/texmf/tex/latex/latex-git/lyluatex/lyluatex.lua:85: in function 'l
ilypond_fragment'
\env@compilerly@process ...g {\localstaffsize } )}
                                                  \env@ignore 
l.12 \end{ly}
           
? 
```

Is this the expected behaviour, i.e. the formatted helpful message being only a "wrapper" around the stacktrace report? Or does this indicate an error in my code?

uliska:
I have pushed that change.

Although it raises an error `err()` allows the overall compilation to continue, and the placeholder is still inserted in the PDF.

However, would it be acceptable to merge that pull request and deal with the option of how to report or not later? (I'd open a separate issue for that). I wouldn't want to work an anything else while this is open because I think the code doesn't allow working on multiple features in parallel currently.

jperon:
The stack trace is the expected behaviour, and that's good : this will let us understand where the error is when a user complains about an error.

uliska:
Thank you

jperon:
Thank *you*: let us not reverse the roles!

Issue #21, opened 2018-01-15
Add option to specify include paths


uliska:
In `compiler_ly` there is already the notion of LilyPond's `-I` command line option, but it only makes use of the file path of the LilyPond file's directory.

There should be the possibility to specify a *list* of include paths, probably on a document-wide level (I'm not sure if it's necessary to change them along the way or even locally).

So the `compiler_ly` function should be modified to iterate over the include argument as a list, such a list has to be made available as a variable in the Lua file, and there should be a macro to set/define this list.

In addition I suggest to also add the directory of the *LaTeX* file to the include directories because it may be an option for many use cases to have include files in *that* directory. This should also be added to the command line when compiling *fragments*.

Issue #33, opened 2018-01-17
Index of scores


uliska:
While storing the scores by their content hash is of course indispensable it effectively renders the tmp_ly directory a black box.

Maybe there should be an index (in addition to the .list file already exported) that allows mapping some reference from the source to the hash.

A suggestion would be to combine that with a label. A key "label" would generate a label after the image and would add an entry to the index

jperon:
> A suggestion would be to combine that with a label. A key "label" would generate a label after the image and would add an entry to the index

I think it would be better to distinguish two things : 

- mapping names to hash, which might be useful for debugging purposes, even without an index;
- table of content or index.

The second point, IMHO, is already managed by ordinary LaTeX commands, and users could wrap them in their own commands; so I'd rather:

- or let users define their own commands;
- or implement this in musicexamples.

jperon:
Do you think the actual behaviour (mapping hash to name in `jobname.list`) is enough, or should we let this open for now ?

jperon:
Closing; feel free to re-open if I forgot something.

PR #40, opened 2018-01-22
Includepaths option

jperon:
To address #21, t's now possible to include files from custom paths by three means :

- `includepaths={path,otherpath}` package option ;
- `\lilypondIncludePaths{path,otherpath}` command ;
- `includepaths={path,otherpath}` option for usual commands.

`{ }` are mandatory only if there are several paths.
There mustn't be any whitespace after the comma.
All paths are relative to current dir.

Even though I don't think `\lilypondIncludePaths` and `includepaths` command option are that useful, they are so similar to `\lilypondCmd` that they were really easy to implement.

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -11,30 +11,50 @@ local err, warn, info, log = luatexbase.provides_module({
 
 local md5 = require 'md5'
 
+
 function ly_define_program(lilypond)
     if lilypond then LILYPOND = lilypond end
 end
 
 
-function flatten_content(ly_code)
+function ly_define_includepaths(includepaths)
+    if includepaths then INCLUDEPATHS = includepaths:explode(',') end
+end
+
+
+function flatten_content(ly_code, input_file)
   --[[ Produce a flattend string from the original content,
        including referenced files (if they can be opened.
        Other files (from LilyPond's include path) are considered
        irrelevant for the purpose of a hashsum.) --]]
-    local result =""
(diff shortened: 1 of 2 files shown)


jperon:
Just added a commit to address #33.

uliska:
@jperon I can't review the code right now as I have to leave for a while, but I wanted to let you know that we may run into merge (and function) conflicts because I have started rewirking the option handling (see my commits on https://github.com/jperon/lyluatex/commits/option-handling - I wouldn't have pushed them already but do so for sake of this comment). 

The idea is that we are going to handle a lot more options than originally thought of, so I think it will be much more manageable to have them handled with an `OPTIONS` and an `LOCAL_OPTIONS` table.
Instead of passing lots of arguments to the Lua functions only the tables are modified if necessary, and the Lua code directly reads from them when necessary.

I have the impression it will work out pretty well, although I still have an issue with option types (can only pass them around as strings).

However, I suggest you continue with your branch (and I'll have a look at it) and we look into that issue when both branches are ready.

jperon:
OK. So I'm going to finish the work (if necessary) on option-handling, then modify my work to make it compatible with it.
Thank you !

uliska:
It *is* necessary as I'm only half through with handling the existing options. I can also finish that, I just have to take a few hours off.

jperon:
No problem for me to do that, and so I'll be able to finish with #21 and #33.

PR #41, opened 2018-01-22
Option handling

jperon:
Here you are; I did it so that options with their default values are set on one statement.


uliska:
Thank you, that looks very neat!
I didn't do much testing with different settings, but it looks like it works.

What I like about your changes is the way local options are handled when they are not given. And that it continues in the direction of leaving TeX as soon as possible and leave the "programming" to Lua. I think that's what LuaTeX actually is about.

I've added one commit completing something I had left lingering this afternoon, and I suggest you merge the pull request.

PR #42, opened 2018-01-22
Includepaths

### LINKED ISSUE #21: Add option to specify include paths

jperon:
This PR is based on #41, which should be reviewed and merged first.
It should fix #21 and #33.

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -38,10 +38,9 @@ function extract_unit(input)
     return {['n'] = input:match('%d+'), ['u'] = input:match('%a+')}
 end
 
-function extract_size_arguments(line_width, staffsize)
+function extract_size_arguments(line_width)
     line_width = extract_unit(line_width)
-    staffsize = calc_staffsize(staffsize)
-    return line_width, staffsize
+    return line_width
 end
 
 function hash_output_filename(ly_code, line_width, staffsize)
@@ -71,7 +70,8 @@ end
 
 
 function lilypond_fragment(ly_code, line_width, staffsize)
-    line_width, staffsize = extract_size_arguments(line_width, staffsize)
+  staffsize = calc_staffsize(get_local_option('staffsize'))
+  line_width = extract_size_arguments(line_width)
     ly_code = ly_code:gsub('\\par ', '\n'):gsub('\\([^%s]*) %-([^%s])', '\\%1-%2')
     local output = hash_output_filename(ly_code, line_width, staffsize)
     local new_score = not is_compiled(output)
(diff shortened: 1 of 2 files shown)


uliska:
I must admit it is too late now to really dig through the code tonight, but I have two comments so far:

1)
Is it OK to require "no spaces around the comma"? OK, in the TeX world people are probably used to that kind of rules, but in general I think we could offer something more permissive. Shouldn't be  too hard.

2)
The include paths should not necessarily be relative to the current directory. Actually this would be the exception rather than the rule. The idea about the include paths is not to shorten the actual include commands but to allow the inclusion of files that can *not* be found relative, for example library files.
I'm not sure if it's necessary to actually integrate files that are included through an absolute include path in the hash consideration (this may include large amounts of rather static library code), but maybe behaviour should be consistent between the flatten function and the code in `run_lilypond()`.

jperon:
It should now be possible to use absolute paths, and a space after the comma is allowed.

There is only one drawback : I don't know why, but the `--include` argument of LilyPond doesn't seem to accept relative paths. So relative paths have to be preceded by `./`, and `lyluatex.sty` replaces it with the full path of current directory.

uliska:
> the --include argument of LilyPond doesn't seem to accept relative paths

Ah really? I'll have to test that.
Your solution seems sound, though, and I'll review ASAP this afternoon.

jperon:
A question that would need to be investigated is the compatibility of all that with Windows ; but I think this shouldn't block merge for now, and could be seen later.

uliska:
Yes, we should do our best to support Windows equally (although personally I don't care about that). But recent investigations with openLilyLib indicate that this isn't too hard since LilyPond seems to convert everything to forward slashes and works well with that (the only notable difference being Windows' volume letter vs. Unix's root directory.

jperon:
That's funny : when I try it from the lua code (removing the `lfs.currentdir()` stuff), this doesn't work. Never mind, as is, it's possible to specify either absolute or relative (with `./`) paths, so let's merge.

Issue #24, opened 2018-01-15
Is there a proper way to purge the temp directory?


uliska:
the lilypond temp directory will grow really fast, with a full set of files for any modification of any score (even size changes). There should be a nice way to purge the temp directory, although it may be difficult to decide what to throw out.

*One* idea (that could be implemented as a package option `purge`) would be to remove all files from the temp directory that are not actually used by the document. For this to work all generated hash filenames should be stored in a list (can conveniently be done in the new function `hash_output_filename` introduced in https://github.com/jperon/lyluatex/pull/18/commits/923a28a69ce72546352f3ae428729671113c7770) and at the end of the LaTeX run all files from the temp directory that don't belong to any of the hashes are removed.

Of course this could trigger unnecessary LilyPond runs, for example when the line width is temporarily modified and later reverted. But that's why it would be a package option and not built in.

jperon:
We could have such an option ; another one would be just to write the list, and to document it so that users may use third-party tools to remove unused files (from a Makefile or such).

rpspringuel:
I just want to chime in and say that having a well defined temporary directory makes it easier on existing purge functions, such as those which might be present in an editor.  I have a feature request which I will be reiterating in about a month (the developer is too busy at the moment to work on it) to allow TeXShop to delete temporary directories when it deletes auxiliary files.  Being able to easily identify the appropriate directory to delete is critical to making such a feature usable.

uliska:
I would say this responsibility may well be left to the user (to use a given default temp directory or select their own). In addition different editors may have different "thoughts" about default directories ...

> We could have such an option ; another one would be just to write the list, and to document it so that users may use third-party tools to remove unused files (from a Makefile or such).

This is of course a good idea. In that case I would vote for *both*: write a list with written/used files (that may or may not become also useful for versioning strategies) *and* offer the option to purge the temp directory. Maybe writing a list could be done always (or at least be on by default) while `purge` (or `purge-ly-dir`) could be a package option (i.e. off by default).

jperon:
This issue is partially solved by 33eca91e9e0a123a97227a2338447886b086474a :

- a `tmpdir` package options lets one define another temporary directory ;
- a `FILENAME.list` is written within the temporary directory.

For now, `FILENAME.list` contains only the hash of the content, which I think should be enough to write custom scripts to clean that directory.

uliska:
[...]

Issue #29, opened 2018-01-17
Also enforce layout of fullpage scores


uliska:
If I'm not mistaken when a LilyPond file is included and `fullpage` is requested it is compiled as defined in the file itself.

However, we should also set the margins for the score according the the geometry of the text document. I think for this we should split `lilypond_fragment_header` into two separate functions, one injecting the system-by-system compilation behaviour (for fragments) and the other defining the `\paper` block, which would also be used for fullpage scores. I have checked that it is possible to append a paper block and staffsize settings to the end of a LilyPond file to override any earlier settings.

Defining the layout of fullpage scores is more involved, as the line width isn't the only relevant parameter anymore, we also have the four margins. And there should also be the differentiation between one- and two-page layout. Probably we should also try to figure out if the full-page score will start on an odd or even page and set the "first-page-number" accordingly.

However, it should be considered/discussed how inserting fullpage scores relates to the corresponding command in `musicexamples`. With all the fragment commands it is clear: they can simply be *wrapped* by the musicexamples environments, but this is a somewhat different case as (so far) musicexamples implements this as a *command*, not an *environment*.

uliska:
I think this is now quite important. With every new option we implement fragments and fullpage examples drift further away from each other.

jperon:
I think I'm going to work on #46, and that those two issues are related. Would you like me to take hand on this one too ?

But I think that, before beginning that, we have to merge existing pull requests.

uliska:
>  think I'm going to work on #46, and that those two issues are related. Would you like me to take hand on this one too ?

That would be nice, I don't have any specific urge to do that myself.

> But I think that, before beginning that, we have to merge existing pull requests.

Yes, definitely. The nature of our code makes merge conflicts far more probable than average.

uliska:
Closed via #48

jperon:
I'd let this one open, as the layout isn't "fully" enforced (margins).

jperon:
#52 advances on solving this issue.

uliska:
Even with #52 merged I'd like to keep this one open as there is yet another issue that may improve the layout results.

What we're aiming at is matching the type area of inserted fullpage scores with that of the text document.

What we have achieved so far is

* match the four `margin` properties
* take "protrusion" into account. While this hasn't fully been resolved (that's the next thing I'll turn back to) we have a good understanding of what we're talking about

What is completely open is the behaviour of the vertical margins.
[...]

PR #43, opened 2018-01-23
`cleantmp` option

jperon:
If `cleantmp=true` is given as a package option, the temporary files (within `tmp_ly`)
that aren't listed in `jobname.list` will be deleted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,10 +3,9 @@ test:
 
 manual:
 	pandoc --pdf-engine=lualatex \
-				 -V fontfamily=libertine \
-		     -o lyluatex.pdf \
-				 lyluatex.md \
-				 && xdg-open lyluatex.pdf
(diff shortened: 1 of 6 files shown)


jperon:
This should fix #24.

uliska:
I think we should not read `FILELIST` but all `*.list` files found in the temp directory.
If we have two tex files in our directory (which share the temp directory) we don't want to remove the legitimate scores from the other document. Or am I missing something?

jperon:
Here you are ; I think it should be ready for merge.

jperon:
Merging now to avoid conflicts when working on #29 and #46. Feel free to open an issue if you see anything bad.

uliska:
Everything seems to work fine (had to scratch my hair a lot until finding out that I should actually *use the option* to test the code ;-)

PR #44, opened 2018-01-24
Fonts to lilypond

### LINKED ISSUE #32: Set fonts in LilyPond to the same as in LaTeX
It would be nice if it was the default that the fonts in use in the LaTeX document (default fonts, the currently active ones?) are also used in LilyPond.

uliska:
Fixes #32

Document fonts can now be injected into the LilyPond score:

* By default the *current* font will become the roman font in LilyPond
* but it can be set so the three families are passed directly
* can be switched on and off globally/locally and along the way

**Note:** So far works only for "fragments" as we don't inject *any* layout into included files, see #29

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ test:
 
 manual:
 	pandoc --pdf-engine=lualatex \
+				 -V fontfamily=libertine \
 		     -o lyluatex.pdf \
 				 lyluatex.md \
 				 && xdg-open lyluatex.pdf
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,12 @@
 test:
 	lualatex -interaction=nonstopmode -shell-escape test.tex
 
+manual:
+	pandoc --pdf-engine=lualatex \
+		     -o lyluatex.pdf \
+				 lyluatex.md \
+				 && xdg-open lyluatex.pdf
+
 ctan:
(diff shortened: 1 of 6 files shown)


jperon:
> For some reason this does not use Libertine as main/roman font, only the sf and tt families are actually used.

I think it comes from the (many !) packages that are included by the LaTeX template of Pandoc. If you specify `-V fontfamily=libertine` as an argument to pandoc, it works as expected.

jperon:
Apart of that, I'd just add *Fr* before *Jacques Peron* (in the manual). That said, I think we could merge this branch (as the font in the lilypond score works as expected), and work on the manual later (perhaps on a special branch, if useful).
Any objection ?

uliska:
OK, fine with me. I've just pushed two more commits fixing the two issues.

I think generally we shouldn't use a separate branch for (most of) the manual work but add stuff in smaller branches, related to the individual features. That is, in retrospect we can of course document everything in one branch, but for example regarding the alignment discussion I thought I'd *start* with writing the documentation as this will give us the best base for discussing what we *want* to have.

jperon:
OK for that. So I merge it, then we could have an `update-doc` branch to make it up with the current state of `lyluatex`, and after that, we'll have to document every new feature while developping it.

PR #48, opened 2018-01-24
Fullpage

jperon:
- Define dimensions for fullpage
- Allow fullpage with fragments


jperon:
N. B. : as you may have noticed, I incidentally replaced all `-` by `_` in hash names : that's because `string.find` behaved oddly with `-`, so instead of losing time, I chose the easy way !

PR #49, opened 2018-01-25
Handling of boolean options (Help needed)

uliska:
I found out that we didn't fully support boolean options (but used them via strings). This means we couldn't do `if get_option('something) then` but `if get_option('something) == 'true' then` - which effectively made *anything* except `'true'` a false value.

8a729333 partially fixes this - but it only works for package options.

With local options I didn't manage to get them to work at all. 3ca8f74c shows my attempts - the culprit is probably how to pass the `keycommand` options along into Lua.

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -358,8 +358,9 @@ end
 
 
 function get_local_option(name)
-    if LOCAL_OPTIONS[name] then
-        return LOCAL_OPTIONS[name]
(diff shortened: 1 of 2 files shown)


uliska:
I started reading the code of #48 and realized that the issue this PR tries to solve may be fixed by that alredy. And if it should *not* be fixed the base should  be quite different.

So this pull request should *not* be considered anymore. I will look into #48 and see if there remains anything to be done.

PR #54, opened 2018-01-29
Fullpage vertical align

uliska:
This Pull Request implements vertical alignment for fullpage scores with two options:

* `fullpagealign=crop`
* `fullpagealign=staffline`
* `\lilypondFullpageAlign{crop/staffline}`

With `crop` the crop box of the LilyPond score is aligned to the typearea box of the text document:
image

With `staffline` the outermost lines of the staff symbol will be aligned, with any additional score items protruding into the margins:
image

I have set `staffline` to be the default because that is consistent how we handle the horizontal protrusion, and because I think that is what we want to have when thinking about beautiful book layout. But we can discuss that.

Probably we should have additional options `extra-top-margin` and `extra-bottom-margin` for manual adjustment for special scores with large protrusion.

### DIFF
--- a/ly/Beethoven/opus-18-1.ly
+++ b/ly/Beethoven/opus-18-1.ly
@@ -7,6 +7,7 @@
 }
 
 \header {
+  title=Beethoven
   % Remove default LilyPond tagline
   tagline = ##f
 }
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -158,27 +158,41 @@ function calc_margins()
     inner-margin = %s\pt
     ]]
 
-    local top = (
+    local tex_top = (
         tex.sp('1in') +
         tex.dimen.voffset +
         tex.dimen.topmargin +
         tex.dimen.headheight +
         tex.dimen.headsep
     )
-    local bottom = (
-        tex.dimen.paperheight - (top + tex.dimen.textheight)
+    local tex_bottom = (
+        tex.dimen.paperheight - (tex_top + tex.dimen.textheight)
     )
     local inner = (
         tex.sp('1in') +
         tex.dimen.oddsidemargin +
         tex.dimen.hoffset
     )
-    return string.format(
-      template,
-      top,
-      bottom,
-      inner
-    )
+    local v_align = get_local_option('fullpagealign')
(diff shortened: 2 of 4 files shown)


jperon:
> I have set staffline to be the default because that is consistent how we handle the horizontal protrusion, and because I think that is what we want to have when thinking about beautiful book layout.

The problem is that gives beautiful layout only when the score has no text above the first staff (neither title, nor anything else). But I think that, most of the time, the main interest of `fullpage` would be to integrate full scores, that would have title, composer etc. typed by LilyPond instead of LaTeX. That's why, IMHO, `crop` should be the default.
But I let you choose the solution you prefer ; feel free to merge whenever you think that's good : I didn't get any error while testing, and won't have much time to merge myself.

uliska:
Did you actually try out with a score with title? In that case `top-markup-spacing` will automatically take precedence over `top-system-spacing`. In fact there is even some extra space which I can't explain (but are willing to accept/ignore):
image

uliska:
I will add the two options for extra margin (which can be usefol in both cases, they can also be negative) and try to figure out if I can find a way to automatically add that tiny bit between the text frame and the actual text (or even the X-height). In the images you can see that there is some whitespace between frame and text in the text side which isn't in the score (where there's actually the horizontal line). *if* we're going into these depths to produce automatic layout I'd like to have that as the final icing on the cake.

jperon:
> Did you actually try out with a score with title?

You might look at, for example, the result of `make test` : the title of Beethoven's score collapses with the line "Allegro con brio", which doesn't occur with `fullpagealign=crop`.

Moreover, this option should be taken in account by `hash_output_filename`, so that the score gets recompiled when it's changed ; but I'm going to open another issue about that and another point.

uliska:
I will look at that example, but I'm sure the problem is not related to *having* the option but to sloppy formatting. These paper layout variables had never been my strong suit, and it seems that once you start touching them you have to define them more completely.

The output filenams was already on my mental list, and I'll fix this tonight (together with some other streamlining of the program flow).

uliska:
Having seen #55 I will *not* try the improvements I had in mind because they would necessarily be less efficient.

PR #59, opened 2018-01-30
Program flow and passing argument

uliska:
This commit (hopefully) streamlines the program flow in the Lua part by not having arguments flow through the function calls (partially they even changed their names during this flow).

Basically the program flow is now like this:

* `lilypond_file()` and `lilypond_fragment()`
These are the entry functions and produce the content LilyPond code from the LaTeX input
* `process_lilypond_code()`
    * First all options are calculated that are needed to determine if the score has to be recompiled.
    * If necessary, `apply_lilypond_header()` is called to set up the Lilypond file/code and `run_lilypond()` to actually compile it
* finally the LaTeX code is generated

Everything that can be considered a "property" of the current score is not passed along as arguments anymore but stored as local options. *Some* values are passed as arguments, and this indicates that they are current values needed to process, not properties of the score.
This may increase the number of calls to `get_local_option()` but I'm sure the clarified flow is more than worth that cost.

This also means that there are *no* ingoing arguments into `hash_output_filename()`, which should be a good base for reworking that function.

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -95,7 +95,7 @@ function write_to_filelist(filename)
     f:close()
 end
 
-function hash_output_filename(ly_code, staffsize)
+function hash_output_filename(ly_code)
     local line_width = get_local_option('line-width')
     local fullpage = get_local_option('fullpage')
     local evenodd = ''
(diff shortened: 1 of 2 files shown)


jperon:
OK for now. I think that I'll make profit of my work on options to modify that in a more "object-oriented" way, and to use the "metatables" of luatex (that should simplify the link between local and global options).

uliska:
One more thing I'd suggest/like to do is some reordering (and potentially renaming) of the functions. As the Lua file grew in such an "organic" way I find the file quite unstructured. I would suggest to either *sort* the functions alphabetically (and then require new functions to e inserted accordingly) or rearrange them in a more natural order reflecting the program flow.

What do you think @jperon ?

jperon:
I totally agree. I'll have time tomorrow afternoon, so I'll take it to seriously work on that, along with some code linting.

Issue #55, opened 2018-01-29
Better handle of options


jperon:
Till now : 
- we define a special command to change every global option within the doc ;
- we add special parts to the name of temporary files in `hash_output_filename` ;
- we define options in three places : at the beginning of `lyluatex.sty`, and when defining `ly` and `lily`.

This should be refactored : 
- we should have only one command, say `\lysetoption{OPTION}`, that would change each parameter ;
- we should take all options in account in `hash_output_filename`, and include them in the hash instead of defining *very* long filenames (that may only get longer while adding options) ;
- we should have a lua function (or another way) to define `ly` and `lily`'s options.

That way, every option could be defined : 
- either globally as a package option ;
- or globally within the document ;
- or locally as a command option.

While it wouldn't necessarily mean something for every option, nothing prohibits to make it possible. Perhaps we would be amazed by users' ingeniousness…

I'll work on this as from wednesday.

uliska:
This totally makes sense. I have just one additional wish: while you're at that could you implement some sore ot (fake) type check in that hypothetical `\lysetoption{}` command? Since we still have strings as input it would be good to check if a given option matches the expected type and/or set of expected values.

uliska:
I *did* start some code clean-up on the https://github.com/jperon/lyluatex/tree/program-flow branch which is intended as a preparation of your options work. It should be mergeable before you start, but if I shouldn't have opened a pull request by the time you want to work on it please check what's going on on that branch.

PR #61, opened 2018-02-01
Code linting

jperon:
For now, we have private functions, a `Score` class and a `ly` public table to expose lua functions to LaTeX commands. Private LaTeX commands are prefixed by `ly@`.
There are quite many things to do yet, but at least the `.sty` is cleaner, and functions in the `.lua` are in a better order.


jperon:
I think #55 should be mostly fixed. Could you tell me what you think, please ?

jperon:
The main purpose of this PR should be OK, I think. If you don't see any mistake, it should be merged; for new functions (error checking etc.), new issues could be opened.

rpspringuel:
Not surprisingly the diffs aren't really informative for this kind of change.  I'm going to have to study the revised code before I can comment intelligently on it and I simply won't have time for that before next weekend (Feb 10 & 11).  If you guys are happy with it before then, don't let my lack of comment stop you from merging.

uliska:
I have read through the code and tested a few of the exmaple files.

The `.sty` file looks very good. The `.lua` file too, but I must say I will need some time to get used to it, so I can't really comment on it. However, I suggest you merge it so we can base any new work on that.

Just one question:

```lua
function Score:set_roman_font()
end
```
in l.449/450, is that a remainder that should be purged or a stub for something to come?

PR #71, opened 2018-02-04
Accept boolean options and dimensions without value

jperon:
The behaviour will be different according on the type of option:
- if the option is boolean, having its value set to '' is equivalent to true
  in lua;
- if it's a dimension, it will be auto-calculated.


uliska:
I have tested this and I think it works.
I'm not sure if we should merge this separately or in one with #74 and #75, though. But I have a question to the ragged-right PR.

PR #74, opened 2018-02-04
Raggedright

jperon:
Please merge #71 first, as this one is based on it.


uliska:
While testing this I can't get `ragged-right=true` to work as a package option. Other combinations work, both as local options and with `\lysetoption`-

uliska:
If I add a (seemingly no-op) `self:raggedright()` at the beginning of `function Score:header()` the package option is interpreted correctly.

I don't really understand how the package option is set, but I have the impression that in the first call of `raggedright()` it is still `default`, regardless of what is set in the .tex document.

Adding that line seems like a workaround, but I would much prefer fixing the underlying issue.

uliska:
Actually it's even more strange. The point was not calling `self:raggedright()` but in having one `print(self['ragged-right']) somewhere in the line.

jperon:
Just corrected.

uliska:
Works now!

PR #75, opened 2018-02-04
Optimize pdf

jperon:
To be merged after #71.


lyluatex.lua
```diff
@@ -409,6 +409,36 @@ function Score:margins()
     end
 end
 
+function Score:optimize_pdf()
+    if self['optimize-pdf'] then
+        local pdf2ps, ps2pdf, pdf2pdf, path
+        for file in lfs.dir(self.tmpdir) do
```
uliska: Maybe I'm not seeing the forest for the trees, but is it necessary to loop over all the files in the temp directory? Wouldn't it be sufficient to write something like "try to open `self.output..'.pdf`"?

jperon:
I think your comment is about #71 instead of this one; but the reply is that you may use `[line-width]`. That makes it behave as in `lilypond-book`.
For example, if you use `[line-width, fullpage]` (not as a package option, as `xkeyval` doesn't allow it), the line width will be calculated (as default), and `fullpage` will be set to `true`.

uliska:
> I think your comment is about #71 instead of this one; 
Yes, but I entered it to the *commit*, so I didn't think it would end up here.

> you may use [line-width]. 
Sounds good, but I'll have to defer testing to tomorrow

lyluatex.lua
jperon: That depends on wether `fullpage` is set or not. I could do this as "try/fail" for each possible name (reading `-systems.count`), but I found it easier to do as is, though perhaps not as "cpu-optimized". I think that it shouldn't be noticed as soon as there aren't thousands of files in the temporary directory.

rpspringuel:
I'm not sure exactly what the purpose of this is.  Is this meant to solve the font subset issue in Lilypond so instead of having multiple copies of the same font with different subsets, there is one font with the union of all the subsets?  If so, then don't you only really need to do it on the final pdf of the document?  Further, wouldn't it also need to be done after LuaTeX is finished making that document?  How does running this pdf->ps->pdf conversion optimize anything when done after each LilyPond run?

uliska:
I tested it, and (apart from taking extra time) it works. The output of compiling `test.tex` is reduced from ~250 to ~150 kB, which is something (especially for larger documents).

One thing I notice, though, is that when compiling with that option, the resulting PDF doesn't show the notation font names anymore, which I think isn't so nice. It's not crucial to have them, but I often look into these document properties to see if everything is right with alternative notation fonts.

@rpspringuel 's questions make sense to me, but I can't really comment.

jperon:
@rpspringuel That's a really good question… I must confess I don't exactlly know wath this command does, but the fact is that with a quite big songbook (549 pages, 77M without any optimization):
- running this after the whole document's compilation gives a 20M pdf;
- running this as done in this PR gives a 13M one.

@uliska The big time needed and the lost font names are why I didn't set it as default option.

rpspringuel:
@jperon That's some pretty substantial size differences.  Do you have the source for that songbook available somewhere so that I can try it myself?

Furthermore, if you don't know exactly what this is doing, where's the background material on which this is based?  Who came up with the idea and in what context?

jperon:
@rpspringuel I know I got this idea from stackexchange or so, but I can't remember where exactly. I wonder wether, instead of subsetting fonts, this command wouldn't outline them, but it's a guess.

The songbook is [here](https://github.com/jperon/sevin); `lyluatex` is called from `lib/mudoc.sty`. You just have to change the `optimize-pdf` option, then `make`.

uliska:
what about investigating a recent change in LilyPond? I recall discussions about that topic and a rather new command line option.
-- 

jperon:
I've heard about it from Knut Petersen; but it's only available in the "cutting-edge" version of lilypond (not even in 2.19), so I think it would be too early to use it already.

uliska:
Ah really? I thought that had been integrated sometime during 2.19.
-- 

jperon:
I'd have to check it.

uliska:
It's totally strange: If I do a web search for "lilypond --bigpdf" I get a first result pointing to http://lilypond.org/doc/v2.19/Documentation/usage/command_002dline-usage. The Google "teaser" reads

> -b, --bigpdfs. PDF files generated will be much larger than normal (due to little or no font optimization). However, if two or more PDF files are included within pdftex , xetex or luatex documents they can then be processed further via ghostscript (merging duplicated font data) resulting in significantly smaller PDF files

which is coherent with my recollection of the discussion. But the actual docs don't contain this. So maybe it has been removed again?

I'll have to look at the Git history when I'm at my proper PC again

rpspringuel:
Check the -O option on the page you linked.  It looks like the option you're remembering got a new name. 

✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝✝
Br. Samuel, OSB
(R. Padraic Springuel)
PAX ☧ ΧΡΙΣΤΟΣ

jperon:
Just tried with lilypond 2.21, using `-O TeX-GS`, and it effectively shrinks size up to 2.7M !
Here are the results:
```
77M     Carnet-no-optimization.pdf
13M     Carnet-pdf2ps-ps2pdf.pdf
3,3M    Carnet-texgs.pdf
2,7M    Carnet-texgs-gs.pdf
```
It could be worth using it

jperon:
After looking at that (and reading again the discussions I had with Knut Petersen), hacking lualatex to run `gs` after finishing the pdf isn't as trivial as it might seem. So I suggest we merge this as is for now, reserving further development on `-O TeX-GS` for later.

jperon:
Are there objections to merge it as is for now ?

uliska:
No, but there should be a new issue collecting the necessary references and thoughts.

rpspringuel:
I haven't had any chance to look at it yet.

PR #77, opened 2018-02-05
Debug lilypond and better handling of lilypond failure

uliska:
This PR includes two features:

* a `debug-lilypond` option that will save both LilyPond's log and the generated LilyPond code in `<output>.ly` and `<output>.log`
* When LilyPond exits with non-zero status a warning is issued, pointing to the log and the .ly file, which are then kept, regardless of the value of `debug-lilypond`.

**Note:** I assume this doesn't work on Windows, but I didn't find a solution for that. (If that's true then it would probably prevent *any* Lilypond call)

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -372,6 +372,29 @@ function Score:lilypond_cmd()
     return cmd.."-o "..self.output.." "..input, mode
 end
 
+function Score:lilypond_version()
+    print("\nCompiling Score with LilyPond executable '"..self.program.."' ...")
+    local p = io.popen(self.program..' --version', 'r')
+    if not p then
+      err([[
+      LilyPond could not be started.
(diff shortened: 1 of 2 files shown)


jperon:
I had several suggestions, but I thought it was easier to propose a commit; feel free to revert it if you don't like it.
Basically, here is what it does:
- change `debug-lilypond` to a more general `debug` option;
- write `.ly` and `.log` only when debugging (I like avoiding unnecessary writes);
- put `lilypond` in silent mode when not debugging;
- ensure "bad" scores are recompiled next time.

jperon:
OK. Thanks !

uliska:
I had already committed dbcb2d92 but not pushed before you merged, so I pushed it directly to master - hoping it is OK.

jperon:
Sure, thank you !

PR #78, opened 2018-02-05
Better handling of compilation failure warnings

uliska:
Follow-up to #77 (but slightly to involved to push directly to master)

With this commit only *one* warning/error is produced,
depending on whether an actual score has been generated or not.
'debug' and 'showfailed' options should work in any combination.

The compilation checks are now concentrated in one function,
un-cluttering run_lilypond() again.

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -224,7 +224,35 @@ function Score:calc_properties()
 end
 
 function Score:check_failed_compilation()
-    if not self:is_compiled() then
+    local debug_msg
+    if self.debug then
+        debug_msg = string.format([[
+        Please check the log file
+        %s
+        and the generated LilyPond code
+        %s
+        ]],
+        self.output..'.log',
+        self.output..'.ly')
+    else
+        debug_msg = [[
+        If you need more information
+        than the above message,
+        please retry with option debug=true.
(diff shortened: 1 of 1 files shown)


lyluatex.lua
```diff
@@ -237,12 +265,19 @@ function Score:check_failed_compilation()
...
+            warn([[
+        LilyPond failed to compile the score.
+%s
+            ]],
+            debug_msg)
             return false
         else
-            err("\nScore failed to compile, please check LilyPond input.\n")
+            err([[
+        Score failed to compile the score.
```
jperon: Wouldn't there be a typo here ? `s/Score/LilyPond/' ?

uliska:
ok, I'll fix and merge when I'm back home.
-- 

uliska:
OK, I've added some more coherence to the reporting and merged the branch. (Unfortunately I deleted the remote branch before pushing the merge, so this PR is labeled as "closed", not "merged".

In addition I added two test file in a new `examples` directory.
I've always tested with lots of random files, and I think it is a good idea to have a directory of example files that (ideally) cover all options and situations. These will not only serve as documentation but also as a sort of regression tests.

PR #79, opened 2018-02-05
Use xkeyval instead of kvoptions

jperon:
This allows using single package options (for example `[cleantmp]`), thus
making package options behave as local options.


jperon:
Merging fast, because I saw another bug to correct, wich would conflict with this.

PR #81, opened 2018-02-06
Discussion: Ignore pdf/eps/ps files

uliska:
PDF, EPS, PS have been commented out in the .gitignore file because they
"might exclude image files for figures".

However, I suggest ignoring them and adding files explicitly if they are
needed for figures etc.


jperon:
That's OK; it was a default .gitignore.

PR #85, opened 2018-02-06
`relative`

### LINKED ISSUE #84: relative
relative=n

Use relative octave mode. By default, notes are specified relative to middle C. The optional integer argument specifies the octave of the starting note, where the default `1` is middle C. `relative` option only works when `fragment` option is set, so `fragment` is automatically implied by relative, regardless of the presence of any `(no)fragment` option in the source.

As best I can tell, the `keycommand` package cannot support a key which is both boolean (given or not given) and general (given with a value).  I'm not sure if other keyword packages (such as `keyval` or `pfgkeys`) provide the needed support.

jperon:
Fix #84

### DIFF
--- a/test.tex
+++ b/test.tex
@@ -1,5 +1,5 @@
 \documentclass[a5paper, fontsize=12pt]{scrbook}
-\usepackage[program=lilypond, tmpdir=tmp_ly]{lyluatex}
+\usepackage[program=lilypond, tmpdir=tmp_ly, pass-fonts]{lyluatex}
 \usepackage{multicol}
 
 \title{}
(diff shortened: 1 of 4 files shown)


test.tex
```diff
@@ -91,14 +89,16 @@
...
+}
 
 Ce fragment :
 
-\lilypond{a' b' c'}
+\lilypond{a' b' c''}
 
 sera écrit plus petit en note de bas de
-page\footnote{\lilypond[fragment]{a' b' c'}}.
+page\footnote{\lilypond[relative=2]{a b c}}.
```
uliska: Just to be clear: We haven't defined any `fragment` option yet, it's just that unknown local options don't raise errors, isn't it?

jperon: For now, really, `fragment` is the default behaviour of `lyluatex` for `ly` and `\lily`. It could be discussed (as it isn't the default behaviour of `lilypond-book`), but it only raises errors when someone tries to use a `\score` block; IMHO, when you need it, it's time to factor out your code to put in within an external file.
That said, I plan to implement `nofragment`, so that if one really wants to do that, he can.

jperon: I'll start working on `nofragment` as soon as this one is merged, to avoid conflicts.

PR #87, opened 2018-02-06
`notime`

### LINKED ISSUE #86: notime
Do not print the time signature, and turns off the timing (time signature, bar lines) in the score.

jperon:
Fix #86


uliska:
... especially the `nostatt` option will come in handy for `inline` fragments.

PR #90, opened 2018-02-06
Lilypond units

jperon:


jperon:
N. B.: as stated in e7a974c,
> This is quite hacky, and IMHO shouldn't be documented, but we support it for
best compatibility with `lilypond-book`.

jperon:
Though a bit convoluted, I've tested enough this to merge, and need this to send other PR.

PR #94, opened 2018-02-06
Nofragment

jperon:
We have a difference with `lilypond-book`: `fragment` is the default. I think we should keep it so, as the main interest of `fragment` (IMHO) is to shorten typing for small extracts. If anyone really wants a `score` block in `\lily` or `ly`, typing `nofragment` won't be such a hassle.


jperon:
Should be fixed now. Incidentally, I added a check that automatically switches off `relative` and `fragment` when there is a `score` block: this will ensure better compatibility with `lilypond-book`.

uliska:
There was one more check that had to be applied, but now it seems good to go (apart from the indicated merge conflicts)

PR #95, opened 2018-02-06
Indent

jperon:
@uliska, there is a problem with this: a very short snippet with `indent=` set will go into the margin. I think this is related to protrusion.
Example: `\lily[indent=10mm]{a b c}` will show the problem.


lyluatex.lua
```diff
@@ -497,8 +496,7 @@ function Score:header()
         )
     elseif self.insert == 'systems' then
 	header = header:gsub(
-	    [[<<<PREAMBLE>>>]], [[\include "lilypond-book-preamble.ly"]]):gsub(
-	    [[<<<PAPER>>>]], [[indent = 0\mm]])
```
jperon: I don't understand this change: in `lilypond-book`, `\noindent` is the default, but it doesn't have a `fullpage` mode. So IMHO, it would be a good thing to have `indent = 0\mm` for `\lily` and `ly`, while auto indent is a good choice for bigger scores included by `\lilypondfile`.

PR #96, opened 2018-02-06
WIP: verbatim

### LINKED ISSUE #91: verbatim
The argument of a LilyPond command is copied to the output file and enclosed in a verbatim block

---

From the docs:

> The argument of a LilyPond command is copied to the output file and enclosed in a verbatim block, followed by any text given with the intertext option (not implemented yet); then the actual music is displayed. This option does not work well with \lilypond{} if it is part of a paragraph.
>
> If verbatim is used in a lilypondfile command, it is possible to enclose verbatim only a part of the source file. If the source file contain a comment containing ‘begin verbatim’ (without quotes), quoting the source in the verbatim block will start after the last occurrence of such a comment; similarly, quoting the source verbatim will stop just before the first occurrence of a comment containing ‘end verbatim’, if there is any. In the following source file example, the music will be interpreted in relative mode, but the verbatim quote will not show the relative block, i.e.

* [x] produce verbatim block  
  (partially implemented, but code from fragments is reduced to a single line)
* [x] `intertext`  
  (not implemented in lilypond-book, but we can of course do it)  
* [x] `begin verbatim`  
  This is *probably* implemented in e3cd2951 but I'm not 100% sure  
  We *only* look for the exact match `% begin verbatim` and not *any* `begin verbatim` in a comment.  
  At least *any* amount of whitespace between the `%` and the `begin` is matched.
* [x] If possible at all it should be enhanced by providing syntax highlighting  
  I think about detecting a loaded package such as lstlistings or minted.  
  Optionally (long-term) we could have a package that interacts with `python-ly` to produce "real" highlighting (better than what Pygments knows about LilyPond). We should then borrow from Pandoc's highlighting (`Shaded` environment).  
For my personal use-case I must admit I'd be fine with *not* implementing this in `lyluatex` but take care of that in a Pandoc filter. This would *not* pass the verbatim option to LaTeX but insert the verbatim block directly because Pandoc already produces fine syntax highlighting.

uliska:
Intends to close #91 

When verbatim is set, print the ly_code as a verbatim block before the score.

**NOTE/QUESTION:**
input from fragments is printed on *one* line, linebreaks are suppressed.
It seems this is already the case when `\BODY` is passed from `ly@compilely`.

@jperon is there a way to pass the content of `lilypond` command/environment into Lua while preserving the newlines?

### DIFF
--- a/examples/verbatim.tex
+++ b/examples/verbatim.tex
@@ -1,0 +1,38 @@
+\documentclass{scrartcl}
+\usepackage{libertine}
+\usepackage[verbatim,
+intertext={The “intertext” option can be used to print some text between the code and the score}]{lyluatex}
+\begin{document}
+
+\section*{Print LilyPond code}
+
+The \texttt{verbatim} option causes the LilyPond code to be printed before the score.  The current issue is that the content of \texttt{\textbackslash lilypond} and the \texttt{lilypond} environment are passed as one line.
+
+\lilypond[intertext=]{ c d e d }
+
+\hrule
+
+\textbf{NOTE:} The local \texttt{intertext} option doesn't work properly yet: it will only work when set to an empty string or to a single word. Setting it to a string enclosed in curly braces or quotation marks will make the option ignored.
+
+\begin{lilypond}[intertext={This change of content does not have any effect}]
+{
+c d e d
+}
+\end{lilypond}
+
+\hrule
+
+\medskip
+\renewcommand{\lyIntertext}[1]{
+
+\textcolor{magenta}{#1}
+
+\bigskip
+}
+The contents of included \emph{files} already work properly:
+
+\lysetoption{intertext}{The formatting of the intertext be adjusted by renewing the command “lyIntertext”}
+
+\lilypondfile[intertext={This change of content does not have any effect}]{../ly/Sphinx/Sphinx}
+
+\end{document}
--- a/ly/Sphinx/Sphinx.ly
+++ b/ly/Sphinx/Sphinx.ly
@@ -1,0 +1,13 @@
+\version "2.18.2"
+
+\header {
+  title = "Sphinxes. No. 1"
+}
+
+{
+  \time 4/2
+  \clef bass
+  as,\breve
+  c
+  b,
+}
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -206,6 +206,7 @@ function Score:new(ly_code, options, input_file)
 end
 
 function Score:calc_properties()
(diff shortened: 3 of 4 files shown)


jperon:
> Is there a way to pass the content of lilypond command/environment into Lua while preserving the newlines?

How a good question…

uliska:
@jperon I sort-of added the `intertext` option but need help.

Obviously I can't set a *local* option to an arbitrary string. Any value except a single word or an empty string will be ignored (that is, the global option will not be overridden). In the package option I can use a long phrase, and I can replace the global option with `\lysetvalue`.

In addition, there are limitations to the possibilities what can be in the string (even if set globally). I noticed:

* a single quote (as in `don't`) will fail because it will be interpreted as the end of the string
* any macros fail (so there's no option to apply any formatting to the intertext)

See the example file `examples/verbatim.tex` for some hints about the issues.

jperon:
> Is there a way to pass the content of lilypond command/environment into Lua while preserving the newlines?
For the environment, it turned to be easier than I feared. The problem is it makes it "unwrappable" in another environment, so perhaps you'll better revert my last commit…

uliska:
I must admit I don't see the difference between https://github.com/jperon/lyluatex/pull/96/commits/a752ac8f361a834f02eec0d16216585bd3cb8b2a and https://github.com/jperon/lyluatex/pull/96/commits/6286fdb3920078ed3a85d5609fef3fe3c4391f7e

So far line breaks are preserved in the `lilypond` environment but not in the `\lilypond` command (which is probably clear but I wanted to state explicitly anyway).
What is *not* preserved yet is *leading* whitespace, i.e. code indentation (which seems plausible since LaTeX ignores that in general).

I think we can live with the limitation that `\lilypond` loses line breaks because that command is intended for single line fragments anyway, IISC.

A minor issue with `\lilypond` is that the enclosing braces aren't printed anymore (which I can fix myself).

So what is still open is the leading whitespace issue. And I still have no clue about it. If you (@jperon) don't have an easy solution I'd look on stackexchange.

jperon:
I'm investigating on it; perhaps I have a beginning solution, but it's not enough clear in my head yet to explain it !

jperon:
I'm slowly coming to a solution, but:
- it will change quite radically the way `lyluatex` passes `ly` content to `lua`;
- I fear it will need to abandon to implement it in a different way.

But I'll keep you informed, as it's quite hard for now !

jperon:
@uliska Is this OK to delete this branch, or would you like to use it ?

PR #100, opened 2018-02-06
`\musicxmlfile`

### LINKED ISSUE #99: `\musicxmlfile{}`
command to insert a MusicXML file, which will be processed by musicxml2ly and lilypond.

jperon:
Though I'm not convinced of its usefulness, that's for compatibility with
`lilypond-book`.
Fix #99


lyluatex.lua
```diff
@@ -122,7 +132,7 @@ local function locate(file, includepaths)
         end
     end
     if not lfs.isfile(result) then result = kpse.find_file(file) end
-    if not result and file:sub(-3) ~= '.ly' then return locate(file..'.ly', includepaths) end
+    if not result and file:sub(-3) ~= ext then return locate(file..ext, includepaths) end
```
uliska: This is not sufficient and will not even match the given `.xml` extention (only after changing to `sub(-4)`, so this must be more robust against different file extensions (changing to -4 will of course break compatibility with `.ly` files).

lyluatex.lua
```diff
@@ -804,15 +814,39 @@ end
...
     ly.score = Score:new(i:read('*a'), options, input_file)
     i:close()
 end
 
 
+function ly.file_musicxml(input_file, options)
+    --[[ Here, we only take in account global option includepaths,
+    as it really doesn't mean anything as a local option. ]]
+    local filename = input_file
+    input_file = locate(input_file, Score.includepaths, '.xml')
```
uliska: Hm, MusicXML files have a number of typical extensions, including `.mxl` and `.musicxml`.
I'm not sure how to deal with that. The "right" way would be to make this an option, but I have the impression this is overkill.

lyluatex.lua
uliska: Actually I'd say it should be checked if `file` does have an extension in the first place and then if it matches `ext`.

lyluatex.lua
jperon: Those extensions (`.ly`, `.xml`) are only a fallback in case the user didn't explicitly specify the extension. so if someone types:
```tex
\lilypondfile{music.ily}
```
it will work; but if he types:
```tex
\lilypondfile{music}
```
it will look not only for `music`, but also for `music.ly`.
So I only took in account the default extension for this, as the user can type other extensions explicitly.

lyluatex.lua
jperon: It *is* checked.

jperon:
I corrected the problem with the extension detection.
About the failure from `musicxml2ly`, yes, that's why I'm not really convinced by this command: IMHO, `musicxml2ly` needs manual checking before inclusion.
That said, my tests with `musicxml2ly` v 2.21 where far better (though far from perfect) than with v 2.18: that sounds quite promising.

uliska:
Thanks. I have the impression the file is now found properly.

But I have experienced another issue. I've randomly downloaded the Fauré sample from http://www.musicxml.com/music-in-musicxml/example-set/ and tried that. In this case `musicxml2ly` crashed when called from `lyluatex` while it works without problems from Frescobaldi. The log output gives a hint:

```
musicxml2ly: Output to `Standard output'
Traceback (most recent call last):
  File "/home/uliska/lilypond/usr/bin/musicxml2ly", line 3277, in ?
    main()
  File "/home/uliska/lilypond/usr/bin/musicxml2ly", line 3271, in main
    voices = convert(filename, options)
  File "/home/uliska/lilypond/usr/bin/musicxml2ly", line 3183, in convert
    score_information.print_ly(printer)
  File "/home/uliska/lilypond/usr/share/lilypond/current/python/musicexp.py", line 855, in print_ly
    self.format_header_strings(k, v, printer)
  File "/home/uliska/lilypond/usr/share/lilypond/current/python/musicexp.py", line 847, in format_header_strings
    printer.dump(value)
  File "/home/uliska/lilypond/usr/share/lilypond/current/python/musicexp.py", line 141, in dump
    self.add_word (w)
  File "/home/uliska/lilypond/usr/share/lilypond/current/python/musicexp.py", line 114, in add_word
    self.newline()
  File "/home/uliska/lilypond/usr/share/lilypond/current/python/musicexp.py", line 123, in newline
    self._file.write (self._line + '\n')
  File "/home/uliska/lilypond/usr/lib/python2.4/codecs.py", line 178, in write
    data, consumed = self.encode(object, self.errors)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xa9 in position 26: unexpected code byte
```

I have always noticed (but so far ignored) that the German messages from LilyPond had encoding problems, but it seems that in this case these actually cause Python to choke, leaving a broken LilyPond file/code behind. It seems there's an encoding issue in the pipe between musicxml2ly and `io.popen()` (I don't think it's between that and LilyPond).

uliska:
A workaround might be trying to write to a file instead of piping, but I assume that's not what you'd want to do. Especially since at that moment we don't have an output filename yet.

jperon:
Yes, and moreover I think it would only hide something that should be reported upstream. Wouldn't this come from the very old version of Python (2.4) ?

uliska:
Oh, that may be true. There has been one user who has several times tried volunteering to update the Python version, but he has always been talked out of it ...

uliska:
But OTOH, I also see encoding issues (just no crashes) with LilyPond itself when called through `io.popen()`: non-ASCII characters are not printed properly. So there may also be something in Lua's libraries, either a bug or a configuration we haven't used yet.

jperon:
Would you mind investigating on it ? I'm currently struggling to solve the problem with `verbatim`, and it would be quite difficult to change now…

uliska:
I'll try - although I must say I've always struggled with this type of problems ;-)

uliska:
It's difficult, so far I've only got the impression that UTF8 support is not really built into Lua.

But what I could confirm is that the output of musicxml2ly crashes on the copyright symbol. Interestingly `debug` actually produces a LilyPond file (because it receives data until the crash), and this includes everything up to the first occurence of the copyright symbol).

I'm not sure who is to blame here, the (old) Python or Lua, but it has to be stated that converting into Frescobaldi does work.

rpspringuel:
Have you looked at the `slnunicode` library in Lua?  Some quick Google searching turned up some potential problems with the `string` library in Lua when dealing with unicode and pointers to this library (which is built into LuaTeX) as the appropriate solution.

jperon:
@uliska, I've tried on my computer with `FaurReveSample.mxl`, and it perfectly works. As my `musicxml2ly` is packaged by Debian, thus using the default `python 2.7`, I guess the problem really comes from `python 2.4`.

uliska:
@jperon I use a downloaded installation which installs wrapper scripts in `~/bin`. It turns out that this script indeed invokes the python executable that is shipped with LilyPodn (how un-UNIX-y ...).

For a test I changed this to run `python` instead, and now the conversion worked without issues. However, there were some syntax errors in the generated LilyPond code, which presumably is "simply" a `musicxml2ly` bug.

How should we proceed?
I'm as reluctant as you spending significant work on this as in general it doesn't even make sense to use the script that way.
Probably we should document that the command has just been added for lilypond-book compatibility and that (and why) we discourage its use.

But there's one thing we should actually do: making the invocation work consistently with the `program` option. `musicxml2ly` can only be called directly the way you currently do it if `program` is also `lilypond`. If there's a path to the executable then the corresponding `musicxml2ly` should be called, it should be in the same directory as `lilypond`.

uliska:
Ah, I have to revert that: you also can't simply invoke it like that. I assume the stuff from the wrapper file is needed too:

```bash
# musicxml2ly wrapper file
#!/bin/sh
export PYTHONPATH="/home/uliska/lilypond/usr/lib/lilypond/current/python:/home/uliska/lilypond/usr/share/lilypond/current/python:$PYTHONPATH"
export GUILE_LOAD_PATH="/home/uliska/lilypond/usr/share/lilypond/current"
export LD_LIBRARY_PATH="/home/uliska/lilypond/usr/lib:$LD_LIBRARY_PATH"
me=`basename $0`
exec "/home/uliska/lilypond/usr/bin/python"  "/home/uliska/lilypond/usr/bin/$me" "$@"
```
I'm not sure this is actually worth it.

jperon:
> Ah, I have to revert that: you also can't simply invoke it like that.

Why not ? The wrapper file being `musicxml2ly` itself, it will define all that stuff. It's up to the user to point to the right executable.

uliska:
I've tested this with LilyPond 2.19.60 and 2.19.80. In both cases the conversion worked but produced faulty LilyPond code.

This was the same when importing from Frescobaldi, so it's definitely not related to the PR but to `musicxml2ly` failing to convert that sample document.

So IMO this is good to go. I'll document accordingly in the `update-manual` branch.

jperon:
I'm on the `lilypond-users` list…

PR #101, opened 2018-02-06
Better handling of <<<PAPER>>> substitution

uliska:
<<<PAPER>>> is not only substituted once, avoiding having two
separate "indent" clauses in the generated LilyPond code.

<<<PAPER>>> is used for the paper size in fullpage mode and set to ''
in the other modes
<<<INDENT>>> is substituted by the result of ly_indent(),
which now takes into account the presence of 'indent' and 'insert=fullpage'


uliska:
I think *this* is now how it should be ...

PR #104, opened 2018-02-07
Process input buffer

jperon:
- better way to capture the `ly` body;
- #91.

I didn't implement `\intertext`. @uliska, I think we should merge this as is for now, but leave #91 open to discuss and implement `\intertext` (IMHO, the name isn't good for that option).

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -831,8 +831,6 @@ function ly.declare_package_options(options)
 end
 
 
-ly.env = {}
-ly.env_name = ''
 ly.score_content = {}
 function ly.env_begin(envs)
     function ly.process_buffer(line)
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -722,7 +722,7 @@ end
 
 function Score:write_tex(do_compile)
     if self.verbatim then
-        ly.verbprint(self.orig_ly_code:explode('\n'))
+        ly.verbprint(self.orig_ly_code:explode('\n'), self['verbatim-highlight'])
     end
     if do_compile then
         if not self:check_failed_compilation() then return end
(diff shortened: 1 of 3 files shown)


uliska:
It's too late tonight for me to read and test your code.

But I'm wondering about the `\intertext` comment, as I didn't use that in my previous attempts.

There is an `intertext` *option* as named in the `lilypond-book` documentation (although it is mentioned as "not implemented"). So I think we don't have a choice but name it like that.

In addition I had `\lyIntertext` as a way to have control over the appearance of the "intertext". This name is open for discussion, but based on the given "intertext".

rpspringuel:
Somebody at some point is going to want syntax highlighting for verbatim code.  Is it even possible for this feature to be made to work with something like the minted package in order to make that happen?

rpspringuel:
Note: I'm not saying we have to implement syntax highlighting ourselves, I'm saying that if someone  takes the time to define an appropriate environment for syntax highlighting LilyPond code, then they should have the option of swapping in that environment for the default `noindent` environment that you're currently using.

jperon:
@uliska Sorry, as it was late this night, I didn't take time to read further about `intertext`. I think we could merge this for now (as there are already quite big changes), then implement it. I really don't find this name ideal (as `quote` and so on), but as it doesn't depend on us, let's keep it.

@rpspringuel Just done as it was fairly easy to implement. I put `\begin{minted}{tex}` till someone implements the `lilypond` lexer for Pygmentize.

uliska:
> I think we could merge this for now (as there are already quite big changes), then implement it.

Fine with me.

>  I put \begin{minted}{tex} till someone implements the lilypond lexer for Pygmentize.

Do you know how Pandoc processes its syntax highlighting? I know that it provides custom macros to format the highlighted code, but it surely doesn't do the highlighting itself? And it *can* highlight LilyPond code.

jperon:
I don't know how Pandoc processes his syntax highlighting, but I'm going to look at this. Thanks for pointing !

lyluatex.sty
```diff
@@ -60,6 +61,12 @@
 }
 \catcode`-=12
 
+\directlua{
+    if ly.get_option('verbatim-highlight') then
+        tex.sprint([[\detokenize{\RequirePackage{minted}}]])
+    end
+}
+
 \newcommand{\ly@setunits}{%
```
uliska: I don't think this is the correct approach because it will fail when the user applies `verbatim-highlight` as a local option only. 

I suggest moving that test to the function where verbatim is actually *used*.
It could be discussed whether it is worse to repeatedly write `\RequirePackage` into the TeX code or testing each time if `minted` is already loaded.

jperon: OK, I'll look at this too.

lyluatex.lua
```diff
@@ -910,4 +946,34 @@ function ly.set_property(k, v)
...
+function ly.verbprint(lines, highlight)
+    local spaces
+    if highlight then
+        --[[ We unfortunately need an external file,
+             as minted is a verbatim environment. ]]
+        local fname = ly.get_option('tmpdir')..'/verb.tex'
+        local f = io.open(fname, 'w')
+        f:write(
+            -- We use tex until one decedes to implement lilypond
+            '\\begin{minted}{tex}\n'..
```
rpspringuel: I'm not sure that forcing `minted` for highlighting is the correct approach.  `pygments` (the highlighter that `minted` is based on) doesn't currently support LilyPond code (as you note in your comment) and there are other syntax highlighting packages (`listings`, for one).  Further, it could be that someone will make use of the functioning syntax highlighter in `python-ly` to write a custom package for highlighting LilyPond code, rather than writing adapting an existing LaTeX package to support a new language.  I was simply using `minted` as an example of a package which can do syntax highlighting.  I think it would be better if we had a internal dummy environment which we alias to `noindent` (or perhaps `verbatim`) by default, but which a user could reassign to a LilyPond syntax highlighting environment defined elsewhere (by whatever method).

This would then nullify the comment below about how and wen to load `minted`.

PR #105, opened 2018-02-08
Add option ly-version

uliska:


uliska:
See 140894b7 for the reasoning

jperon:
Very good idea; I wonder why I didn't do that before !

PR #107, opened 2018-02-08
Let the user define his own `verbatim` environment

jperon:
Following @rpspringuel's suggestion about #104.

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -723,7 +723,7 @@ end
 
 function Score:write_tex(do_compile)
     if self.verbatim then
-        ly.verbprint(self.orig_ly_code:explode('\n'), self['verbatim-highlight'])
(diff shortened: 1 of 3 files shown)


uliska:
Works correctly, but I'm not 100% happy with the interface. Asking the user to actually pass LaTeX code to a macro seems somewhat inconsistent with our usual behaviour.
I'd prefer something like
```tex
\lyhighlighting[language=TeX]{lstlisting}{listings}
```
that will generate the environment *and* load the package in one step.

jperon:
The problem with such an approach would be that only an environment could be used. Perhaps someone could want to use a command before and a command after, not necessarily an environment.

jperon:
The logic in `\lysetverbenv` is the same as `\newenvironment`, so it shouldn't be difficult to understand for the user.

uliska:
OK.

PR #108, opened 2018-02-08
Implement % begin verbatim

uliska:
Check for % begin verbatim parts in code to be printed verbatim.

Hm, this will probably conflict (merge-wise, not content-wise) with #107


jperon:
I'm going to resolve conflicts and merge.

PR #111, opened 2018-02-08
Allow multiple of lenghts as dimension arguments

### LINKED ISSUE #103: "Percent" value
If I would want to implement "percent" as a value for dimensions, where would I go to define it? And how should it be written in the LaTeX input (obviously we can't use `80%`)?

I would like to have to option to use percent for
* `line-width`
* protrusion (use up to 60% of the available margin)
* `staffsize` (80% of the default size)

uliska:
Closes #103

With this commit it is possible to write
```
    [line-width=.7\linewidth]
```
All lengths can be used, for example
```
    [line-width=1\textheight]
```
to produce a square.

The value can take all forms:
* 0.8
* .8
* 2
* *<empty>*

The last means that `[line-width=\textheight]` will be interpreted as `1\textheight`


uliska:
I forgot to mention that LaTeX itself checks the validity of the length (don't know why).
So 
line-width=\something
will fail as undefined command.

PR #117, opened 2018-02-10
Check for current dir only once

uliska:
Done on the mobile browser, so please look closely.
But I think we should only store the current file directory once while loading the package.

### DIFF
--- a/lyluatex.sty
+++ b/lyluatex.sty
@@ -74,6 +74,11 @@
 }
 \catcode`-=12
 
+% Store the current file's directory
+\directlua{
+  ly.CURRFILEDIR = '\currfiledir
+}
+
 \newcommand{\ly@setunits}{%
   \let\ly@old@in\in\protected\def\in{in}%
   \let\ly@old@pt\pt\protected\def\pt{pt}%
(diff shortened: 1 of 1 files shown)


rpspringuel:
If the "current file" in "current file directory" refers to the main project tex file (which is the file being read when the package first loaded), then this change makes sense.  However, if "current file" refers to the file being read when the `\ly@compilescore` command is called, then not so much.

uliska:
The "current file" refers to the (main) .Tex file being compiled and is used to generate file names in the temp dir.
-- 

jperon:
@uliska No, it isn't the main.tex (it wouldn't be useful), but the directory of the .tex from which the score is called (which could be a secondary file called by \input and located in another directory).
Sorry for the silence, I'm far from my computer and writing from my phone.

uliska:
Ah?

So what does that mean for tmpdir and relative LilyPond files? Wouldn't this generate an `output` variable relative to that current file's directory? And that isn't created with mkdirs() at the start?



-- 

uliska:
And: if we check the directory of the current (potentially included) file from which the current score is triggered, why is `ly.CURRFILEDIR` a global variable? Shouldn't it then be a property of the `Score` object?

jperon:
Good remarks; closing this one, but follow-up on #120.

PR #118, opened 2018-02-10
Fix obsolete comment

uliska:
This comment was a leftover from a previous state.

Additional question: is it necessary to do that in the TeX Part?  it seems cleaner to ask for that value directly from Lua, if that's possible.


rpspringuel:
The comment fix is fine.

Examining the sty file, it seems that the page number inquiry is specifically for the Lua code and so could probably be moved over.

jperon:
I didn't find a way to get the current page number directly from lua code. If you know one, feel free to submit a PR, as it certainly would be cleaner than passing it as it is now.

PR #120, opened 2018-02-11
Corrections about includepaths

jperon:


uliska:
I must admit that it's hard for me to follow the changes in this PR, both from reading the diff and the new code.

What should this do exactly, and how should I test it?

jperon:
It should ensure that this works:
in main.tex:
```tex
\input{subir/test}
```

in subdir/test.tex:
```tex
\lilypondfile{score}
```

with a score.ly in subdir.

lyluatex.lua
```diff
@@ -129,13 +131,11 @@ end
...
-            if lfs.isfile(result) then break end
-        end
+    local result
+    for _, d in ipairs(extract_includepaths(includepaths)) do
+        if d:sub(-1) ~= '/' then d = d..'/' end
+        result = d..file
+        if lfs.isfile(result) then break end
     end
     if not lfs.isfile(result) then result = kpse.find_file(file) end
     if not result and ext and file:match('.%w+$') ~= ext then return locate(file..ext, includepaths) end
```
uliska: @jperon It is unrelated to the PR but I don't see what you want to achieve with the match here. If I read this correctly it will match "any character, followed by one or more alphanumerical characters" which will never match the extension but something at the beginning. Shouldn't that rather be `match('%..+)` to match a literal dot followed by any (and not only alphanumerical) characters?

jperon: Right; fixed in #123. It worked as it was (because the `.` wasn't followed by `+`, `*` or `?`, but it's more readable now.

Issue #98, opened 2018-02-06
LilyPond indent fails for single-system scores


uliska:
When a score is indented with the new `indent` option (#95) it protrudes in the left margin *if it is a single-system score*.

The reason is that LilyPond cuts the indented space off by offsetting the bounding box (visible in the *.eps output). With multi-system scores this is not an issue. This offset is interpreted as a negative protrusion and incorrectly "adjusted".

I'll have to somehow calculate this offset against the nominal indent (in `pt`) in order to also deal with actual protrusion

uliska:
Score indentation makes real sense only for multi-system scores where the *first* system is indented (which is actually what the `lilypond-book` docs say).

Wouldn't it be a better option to silently (but of course documented) deactivate `indent` for scores that produce a single system? One could check from the intermediate files and trigger a re-run with changed options.
We will have to implement a control structure for re-running LilyPond anyway, in the context of the "max protrusion" option.

So the only argument against suppressing the indent for single-line scores is that users might be surprised.
Opinions, @jperon @rpspringuel ?

jperon:
I agree with that, but don't think a re-run is necessary: it could be implemented within `Score:write_tex` and `Score:protrusion`. For example, `Score:protrusion` could have a `nsystems` argument, and if it's equal to `1`, take off `self.indent` from protrusion. The user could be warned through `warn()`.

jperon:
Incidentally, this could be an occasion to add a `noprotrusion` option, and perhaps a `protrusion` to let the user force its value.

jperon:
I think this issue should be solved by #123. Is it OK ?

uliska:
There's one limitation where this behaviour will produce suboptimal results, but I'm not sure if this corner case warrants further development (or if it's sufficient to document it).

1. Consider a score that is just a little bit too long to fit in one indented system - but it would fit in a single system if it were not indented. In that case it will stay indented and be printed on two systems.
2. Or if we want to include the first system of a multi-system score it would be better to actually compile an unindented version

Regarding 1. I think one could trigger an unindented re-run if the system count is exactly 2. If the result still has two systems the original score will be used, otherwise the new single-system score.

Regarding 2. it is probably OK to expect the user to disable the indent if they actually want to do that.

I would say we should look for a solution to 1. and keep this issue open but remove the "bug" label and the Release v1 milestone. What do you think?

jperon:
That's OK for either 1. or let it as it is (the user having the option to force noindent in corner cases).

PR #123, opened 2018-02-11
Refactor - fix #116

### LINKED ISSUE #116: Clean up write_tex() function
I think in order to improve the maintainability and especially extensibility of the code we should rewrite the `Score:write_tex()` function (similar to what we did with option handling).

I think that function should *not* deal at all with low-level TeX output but only define the program flow through function calls. This way it will be easier to add future functions that change the way the TeX code is generated (e.g. conditionals). Currently it's somewhat intimidating to start adding a new function that affects the TeX code generation.

I suggest to consistently prefix the names of the functions called in here, e.g. with `Score:tex_`, so these functions will be grouped at one sorting position.

### LINKED ISSUE #124: addversion
(Only for Texinfo output.) Prepend line \version @w{"@version{}"} to verbatim output.

---

*rpspringuel*: Even though this is a texinfo option only in lilypond-book, I think it could be useful in non-texinfo output.

*uliska*: Yes, that may be interesting in other contexts too

jperon:
All functions that write LaTeX code are now in a `latex` array. They need to
receive explicit arguments, so that it's clear where each score property is used.

Also fixes #124

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -12,7 +12,8 @@ local err, warn, info, log = luatexbase.provides_module({
 local md5 = require 'md5'
 local lfs = require 'lfs'
 
-ly = {}
+local ly = {}
+local latex = {}
 
 local FILELIST
 local OPTIONS = {}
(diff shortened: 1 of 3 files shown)


jperon:
As I had to refactor the `Score:protrusion()` function, it was an occasion to fix #98.
N.B.: with `noprotrusion` activated, it's evidently up to the user to manually specify the line-width if he wants to avoid "protrusion" to the right margin…

uliska:
Yes, this is right. There is one situation where not fixing this with a second LilyPond run may cause inferior results: when a score *just* exceeds one system. In that case it will be intended and have two systems while it could have been set on one line if indentation had been switched off before (the second) compilation.
But I think this is a corner case we can safely neglect. It can be documented and the user can fix it with 'noindent'.
-- 

lyluatex.lua
```diff
@@ -216,6 +210,72 @@ local function splitext(str, ext)
...
+end
+
+function latex.label(label, labelprefix)
+    if label then tex.sprint('\\label{'..labelprefix..label..'}%%') end
+end
+
+function latex.verbatim(verbatim, ly_code, intertext, version)
+    if verbatim then
+        ly.verbprint(ly_code:explode('\n'))
+        if version then tex.sprint('\\lyVersion{'..version..'}') end
```
uliska: Is it uncontroversial that the version is printed *after* the verbatim code?

lyluatex.lua
```diff
@@ -770,62 +839,22 @@ function Score:run_lilypond()
...
-            [[\ifx\postLilyPondExample\undefined\else\expandafter\postLilyPondExample\fi]]
-        tex.sprint(texoutput:explode('\n'))
+    latex.fullpagestyle(self.fullpagestyle, self['print-page-number'])
+    latex.label(self.label, self.labelprefix)
+    local systems_file = self.output..'-systems'
+    if not lfs.isfile(systems_file..'.tex') then  -- fullpage score
+        latex.includepdf(self.output)
+    else  -- fragment
+        latex.includesystems(
+            systems_file, self:_protrusion(), convert_unit(self.indent), do_compile
```
uliska: This will fail in line 821 when `do_compile == false` (i.e. when a document is compiled the second time) because the `.eps` file has been removed after the first run.

I have to leave the computer now, therefore I can't  look for a fix.

jperon: It didn't fail, because `_protrusion()` returned `nil` and this result wasn't used in `latex.includesystems()`. But I rewrote it in a more explicit way, and it gave me the occasion to enhance `latex.includesystems()`.

uliska: It *did* (does) fail before `_protrusion()` gets a chance to return because there is no `.eps` file to open, therefore `f` is nil and `f:read()` fails.

uliska: Minimal document:

```TeX
\documentclass{scrartcl}
\usepackage{lyluatex}
\begin{document}

\lilypond{ c' }

\end{document}
```

Failure upon *second* compilation:

```
lyluatex/lyluatex.
lua:821: attempt to index local 'f' (a nil value)
stack traceback:
	...e/uliska/texmf/tex/latex/latex-git/lyluatex/lyluatex.lua:821: in function '
_protrusion'
	...e/uliska/texmf/tex/latex/latex-git/lyluatex/lyluatex.lua:857: in function '
write_latex'
	...e/uliska/texmf/tex/latex/latex-git/lyluatex/lyluatex.lua:809: in function '
process'
                                                  
l.5 \lilypond{ c' }
                 
?
```

jperon: That's weird, I didn't have that error. Never mind, it's fixed now.

uliska: Thanks, works now.

I must admit I'm not 100% I've assessed and tested everything, but I think this PR is good to be merged now.

lyluatex.lua
rpspringuel: I would have expected it before, as that is the convention with LilyPond.  Not that I think the program actually cares.

PR #130, opened 2018-02-12
quote option

### LINKED ISSUE #129: quote
Reduce line length of a music snippet by 2*0.4in and put the output into a quotation block. The value ‘0.4in’ can be controlled with the `exampleindent` option.

---

*uliska:* I find the name `quote` pretty nonsensical, and I wouldn't use a `quote` environment. What we have is a centered score that is indented. The value to be given is the indentation on one side, which the program has to double transparently.

Is it clear how that is handled with regard to protrusion?

jperon:
Fix #129
Though the name `quote` isn't ideal, I put it till we find a better one.
Added also `gutter`, `leftgutter` and `rightgutter`; `exampleindent` is an
alias for `gutter`.


uliska:
I would just keep that name since we are forced to use it. It is not bad enough to justify duplicating it with an alias IMO

lyluatex.lua
```diff
@@ -1013,7 +1031,10 @@ function ly.get_option(opt)
 end
 
 
-function ly.is_dim (k, v)
+function ly.is_alias() end
```
uliska: What is this empty function for?

lyluatex.sty
```diff
@@ -30,9 +30,12 @@
     ['fragment'] = {'true', 'false', ''},
     ['fullpagealign'] = {'staffline', 'crop'},
     ['fullpagestyle'] = {''},
+    ['gutter'] = {'.4in', ly.is_dim},
+        ['exampleindent'] = {'gutter', ly.is_alias},
+        ['leftgutter'] = {'', ly.is_dim}, ['rightgutter'] = {'', ly.is_dim},
```
uliska: I will have to reconsider the use of the gutters when I (finally) get to rethinking about the whold protrusion thing, but for now I'd say this is good to go.

lyluatex.lua
```diff
@@ -301,9 +309,17 @@ function Score:calc_properties()
     if staffsize == 0 then staffsize = fontinfo(font.current()).size/39321.6 end
     self.staffsize = staffsize
     -- dimensions that can be given by LaTeX
-    for _, dimension in pairs({'line-width', 'paperwidth', 'paperheight'}) do
+    for _, dimension in pairs({
+        'gutter', 'leftgutter', 'rightgutter', 'line-width', 'paperwidth', 'paperheight'
```
uliska: I think this table coud be factored out to a `local.DIMOPTIONS` table

lyluatex.lua
jperon: It's to keep the same behaviour than others when declaring it as a package option. It could be an arbitrary variable, providdent that its value isn't "reproducible"; I thought it wasn't harder to set it as a function.

PR #131, opened 2018-02-13
Add option 'print-only' (choose pages/systems to be included)

### LINKED ISSUE #35: Include only selected systems or pages
When having the score encoded in separate files it may be desirable to only print a subset of the score. That way one can have one complete encoding of something and refer to specific points in score examples. Valid requests include (more suggestions welcome):

* The first system (incipit)
* A certain number of systems  
  (“I want to show some amount of music or fill some amount of paper but don't care about the *exact* amount of music”)
* A certain range of measures

The first two would have to be done in the stage of `write_tex()` while for the third `\set score.skipTypesetting` commands would have to be injected in the `ly_code'' somehow. But that might be difficult without adding the [edition-engraver](https://github.com/openlilylib/edition-engraver) as a dependency.

uliska:
Closes #35 
Documentation in commit https://github.com/jperon/lyluatex/commit/15ff50b4af451b59cf2a78b2c8bebc30df95f9a9

'print-only' can filter the systems or pages to be printed,

Value can be
- empty (= all)
- single number
- range (M-N) (N-M is also possible)
- list (A,B,C, D, E ,F) (arbitrary order)

Note that the user is responsible for providing numbers within
the actual range of the score.

If only the first system is included *and* 'indent' is set to a value
then the indent is deactivated


rpspringuel:
What about something like `1-3,5, 9-10` (i.e. a combination of list and range).

uliska:
I thought this wasn't worth the additional effort since it can easily be faked by multiple includes (at least with included files) and you'd usually want some text between the selections anyway.
But now that you forced be to think about it I found a perfectly valid use case: say you have a PDF containing many scores, e.g. a song book, then you can use this option to print arbitrary selections of songs.

Additionally I realized that I also have to ensure this option is *not* considered for the file name hash because we do *not* want different selections cause the score to be recompiled.

So **this PR is not ready to be merged**.

uliska:
I improved the parsing, so now any number, range, or list of number/range items can be used (whitespace in ranges will also be parsed correctly).

Also, changing the items to be included won't trigger recompilation (which is of course important when including *different* excerpts from the same score.

Finally there's a hidden change (at the occasion of rewriting how systems are written): I changed the default `\linebreak` between systems to `\par\bigskip` because this is how consecutive systems should look like. Using simple line breaks should be the absolute exception that can be done explicitly by the user.

**Now I think it's ready to be merged.**

lyluatex.lua
```diff
@@ -230,31 +230,87 @@ function latex.fullpagestyle(style, ppn)
...
+        texoutput = texoutput..string.format([[
+        \noindent\hspace*{%spt}\includegraphics{%s}
+        ]],
+        protrusion, filename..'-'..system)
+        if ly.between_lilypond then
+            texoutput = texoutput..string.format([[
+            \betweenLilyPondSystem{%s}
+            ]], index)
+        else
+            texoutput = texoutput..'\n\\par\\bigskip'
```
jperon: I don't really agree with this: the result is good only when `staffsize` is auto-calculated, not when it's far enough from the default (for example `staffsize=16` with a document's fontsize of 12pt).
I think it should:
- either stick to the default `\linebreak` (as in `lilypond-book`);
- or be a `\vpsace` proportional to `staffsize`.

uliska:
I've replaced `\bigskip` by a flexible space related to *staff*size.

Test with

```TeX
\documentclass{scrartcl}
\usepackage{blindtext}
\usepackage{lyluatex}
\begin{document}


\blindtext

\begin{lilypond}[]
  \repeat unfold 40 { c' d' e' d' }
\end{lilypond}

\blindtext

\begin{lilypond}[staffsize=12]
  \repeat unfold 40 { c' d' e' d' }
\end{lilypond}

\blindtext

\begin{lilypond}[staffsize=16]
  \repeat unfold 40 { c' d' e' d' }
\end{lilypond}

\blindtext

\begin{lilypond}[staffsize=15]
  \repeat unfold 40 { c' d' e' d' }
\end{lilypond}

\blindtext

\begin{lilypond}[staffsize=5,line-width=.4\linewidth]
  \repeat unfold 40 { c' d' e' d' }
\end{lilypond}

\end{document}
```
[between-systems-space.pdf](https://github.com/jperon/lyluatex/files/1720441/between-systems-space.pdf)

PR #133, opened 2018-02-13
insert=inline (WIP)

### LINKED ISSUE #45: Wish: inline scores
I'd like to implement a third mode beside fullpage and fragment: inline. Basically this should produce an image and include it like a character - actually a `lilyglyphs` command but "live". Differences from fragments include

* other boilerplate
* options to hide/show stafflines, clef, key (default: omitted)
* no handling of indentation
* maybe a different staffsize heuristic

uliska:
Closes #45
Is based on #131 and can only be merged after that.

This adds the `insert=inline` mode to insert music *inside* paragraphs, i.e. in the continuous text.  There is one bug, and I have questions regarding the interface, so the branch has to be completed after som discussion.

---

For some reason there is some leading space before the image:

image

I have double-checked with an `\fbox{}` and with manually including the resulting image: the image is properly cropped and the space comes from the inclusion in the document. I had the suspicion that line breaks unprotected by `%` would cause this (as that's the typical reason), but I haven't succeeded in fully removing that space (https://github.com/jperon/lyluatex/commit/219ac9c846923e7961ad61043b27bfcb9fc826c8 made it better, though).

---

I'm not sure how to proceed with the interface for the staffsize.

I think in most of the cases the staffsize of an inline score is significantly smaller than that of regular score examples (the above image has `staffsize=12` in a default text document for example).

I think for the `default` case we can come up with an additional ratio that is applied to inline scores, but how to go about manually set values? I think that for manually setting the options it should be possible to set them once and work properly with both inline and regular scores. I see the following possibilities but would be glad about further suggestions:

* manually set and reset `staffsize` whenever inline and regular scores change (current behaviour, inacceptable)
* define an additional staffsize option (e.g. `inline-staffsize`) to set them independently
* keep `staffsize` as it is, referring to the absolute staffsize of regular scores, and silently apply a given ratio for inline scores. Optionally add an option to control that ratio.

Somehow I'm not totally happy with either option because they feel slightly too complex (from the user's perspective). The third one would provide the "smoothest" interface to me.


jperon:
IMHO for staffsize, the second option `inline-staffsize` would give a more "expected" behavior. It could be set by default to a value proportional to `staffsize`, which would be a compromise with the third option.

uliska:
ok, I'll do that tomorrow
-- 

uliska:
`inline-staffsize` added in f6343dea.

Additional question: should we provide means for automatically improve vertical alignment of inline scores?
Currently I've implemented the `voffset` option that can manually shift the vertical position, but maybe it would be nice to have something better? By default the bottom edge of the image is aligned to the text's baseline, which looks bad in cases like this:

image

Of course I'd love to align (again) the lowest staffline to the text baseline, but I'm not sure that's possible to determine. But at least vertical *centering* should be possible. What do you think?

jperon:
Centering may be easily obtained thanks to the `\pbox` command of `pbox` package.
So you could add `\RequirePackage{pbox}` in `lyluatex.sty`, then modify `latex.includeinline` as follows:
```lua
function latex.includeinline(pdfname, voffset)
    tex.sprint(string.format([[
\raisebox{%s}{\pbox{\linewidth}{\includegraphics{%s-1.pdf}}}
]], voffset, pdfname))
end
```
Perhaps we could make it optional, as you like.

jperon:
The option `bare-inline` uses default `staffsize` and not `inline-staffsize`. Is it intended ?

uliska:
No, I simply forgot that. Have to go out of house now and will continue finishing that later today.

Thanks for the fix with https://github.com/jperon/lyluatex/pull/133/commits/3c6c245d0d3011c4e266d84ac26e6cd7a2ac2409
I confirm that the horizontal spacing is correct now.

uliska:
Am 14. Februar 2018 11:47:10 MEZ schrieb jperon <notifications@github.com>:
>Centering may be easily obtained thanks to the `\pbox` command of
>`pbox` package.
>So you could add `\RequirePackage{pbox}` in `lyluatex.sty`, then modify
>`latex.includeinline` as follows:
>```lua
>function latex.includeinline(pdfname, voffset)
>    tex.sprint(string.format([[
>\raisebox{%s}{\pbox{\linewidth}{\includegraphics{%s-1.pdf}}}
>]], voffset, pdfname))
>end
>```
>Perhaps we could make it optional, as you like.

I will look into this, but I will check if I can get more information from the bounding box which would allow us to have more control.
-- 

uliska:
> Centering may be easily obtained thanks to the \pbox command of pbox package.

The bounding box seems to give information about the vertical extent of the file to some reference point.
I'll investigate this further.

jperon:
After a quick look, I fear the bounding box won't help us much this time: for cases as different as `\lilypond[insert=inline]{a' b' g'''}` and `\lilypond[insert=inline]{a' b' d}, I get very close results:
`0 -32 53 1` and `0 -31 55 0`.

jperon:
So what are we doing: let it as is, or vertically center it ?

uliska:
I've just managed to get back to it: I'll add an option `valign=center|top|bottom` with default `center`. But I'll get it from the bounding box, without an additional package dependency.

uliska:
I think this is ready to be merged now:

* `insert=inline|bare-inline` works
* Automatic staffsize, controllable with `inline-staffsize`
* Vertical alignment with `valign=center|top|bottom`
* Manual adjustment with `voffset`
* Horizontal padding with `hpadding` (default 1ex)
* Check for score with multiple systems

What *can't* be done yet is aligning to the *content*, i.e. the staff lines, but I think this isn't possible.

jperon:
What about the small modification of last commit ?

jperon:
I tried it too, but IMHO it would be a bit too low (by default, i. e. when no note goes below or over the staff).

uliska:
OK, I'll just believe you. So would you now merge this PR?

This makes me think that the next release should be a "beta" release since we *think* we're ready with a basic 1.0 version and lilypond-book compatibility - but it may be really good to have feedback from users from the "real world" before such a statement.

jperon:
That's evident for the *beta*: there has been so much work since 3 weeks that there *should* be bugs…

jperon:
I don't see any fdd14d17 (it doesn't point to a commit in your message).

jperon:
Ah, just seen !

uliska:
I'll try to fix that when I'm ready with the work.
I think I'll first create a temporary branch from `master`, then revert https://github.com/jperon/lyluatex/pull/133/commits/99b15d931ed862f6d17dca721d833c8de5d3157c on that branch, merge master into `caption` and then review your branch and see what should go into the modified code where the arguments are not passed around individually anymore)

Issue #50, opened 2018-01-25
Integration with musicexamples package


uliska:
I will now start reviewing the `musicexamples` package, with three main targets:

* general coding review (I assume there's much to be improved)
* removing obsolete attempts to achieve what `lyluatex` now does
* ensuring compatibility with `lyluatex`

We should try to minimize functionality overlap as much as possible in order to make the two package perfect complements. Actually I even considered *integrating* `musicexamples`'s functionality into `lyluatex` because that would be the most natural thing to do. However, I think `musicexamples` should continue to work with *any* LaTeX engine, so that's not an option.

The general idea should be that `lyluatex` generates an `\includegraphics` expression that can be inserted in a `musicexamples` environment. This will work very easily with fragments, but fullpage scores are a more complex issue (which I'll discuss in a separate issue when the time is there for that (after more review of the other package)).

My idea about the integation is the following: when `lyluatex` knows that `musicexamples` is loaded (this would mean `musicexamples` would have to be loaded first) it changes some parts of its operation and automatically wraps the `\includegraphics` commands in `musicexamples` environments. For the user the immediate difference would be that the scores aren't simply placed wherever they fall but are beautifully spaced by environments. And simply passing a caption in the optional argument will do all the magic of numbering, adding to a list of music examples etc.). This will lead to quite some more new options, but by now that shouldn't scare us anymore.

rpspringuel:
If you make use of the [`afterpackage`](https://ctan.org/pkg/afterpackage) package then there is no need to worry about the package loading order.

jperon:
Now that most developments about `lyluatex` itself are done, here are my thoughts about this point. I really like the reference to `\includegraphics`: IMHO, `lyluatex` should be for `musicexamples` what `\includegraphics` is to floating environments. `\includegraphics` isn't *wrapping* anything, it is *wrappable*. So we should ensure the easiest *wrappable* behavior from `lyluatex`, letting *wrapping* to `musicexamples` (or other packages, as users could have other approaches to this problem).

So I agree that `musicexamples` shouldn't depend on `lyluatex`; but on the other side, I don't think `lyluatex`'s behavior should be impacted by `musicexamples`.
Perhaps the idea that simply adding a caption does all the magic has something appealing, but I also think it may be confusing. IMHO, the following approach would be less confusing:

```TeX
\newenvironment{lyfigure}[2][]{%
  \edef\mycaption{#2}
  \figure
    \center
      \ly[#1]%
}{%
      \endly
[...]

PR #135, opened 2018-02-16
Caption/Environment/horizontal alignment

uliska:
I've mostly undone my refactoring in the commit merging `master` into my working branch. I ran quite a number of tests, but I can't fully see if there are remaining issues (that's the point where we should already have a full suite of regression tests in the `examples` directory ...

Documentation of the behaviour is in bad9423c and db9a2303

I'll only go on to the protrusion issue when this PR is merged.

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -18,8 +18,16 @@ local latex = {}
 local FILELIST
 local OPTIONS = {}
 local DIM_OPTIONS = {
-    'gutter', 'leftgutter', 'rightgutter',
-    'line-width', 'paperwidth', 'paperheight'
+    'extra-bottom-margin',
+    'extra-top-margin',
+    'gutter',
+    'indent',
+    'leftgutter',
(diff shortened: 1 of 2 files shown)


jperon:
Though I didn't have time to inspect it deeply, I see several problems with this PR:
- protrusion is (quite severely) broken with `insert=systems`, especially with `musicxmlfile` (cf. test.tex);
- I don't understand this approach about wrapping within an environment in `lyluatex` itself; but I'm going to discuss it further in #50.

uliska:
I don't see the protrusion as a problem as it has been inherently broken since the `protrusion` option has been added, and the code has to be substantially reviewed anyway. I could do any of

* try to find what causes the issue in the *current* code and fix that before merging
* continue working on the protrusion on *this* branch before merging
* just merge and have a "broken" state for a short time in the Git history

I'll also discuss the wrapping issue on #50 although that is not really appropriate anymore.

PR #142, opened 2018-02-20
Protrusion

### LINKED ISSUE #36: Discuss the alignment of scores to the text area
Now that #31 has been merged `lyluatex` applies protrusion and correctly determines the correct line width. But this raises further questions, on a typographical, technical, and interface level. Attached is a [detailed PDF](https://github.com/jperon/lyluatex/files/1642770/test-alignment.pdf)* demonstrating the issues. The questions to be discussed in order to allow further development are (additions welcome):

* Should protrusion be used always or are there valid reasons to avoid it?
* What is the correct way to deal with the issue (from the typographical perspective)?
* Which interface should be provided for package users?

*(the source code used to produce this can be found as `test-alignment.tex`)

### LINKED ISSUE #39: Also take protrusion to the right into account
Discussion in #36 made me realize that a score can also protrude to the right, and that should be taken into account as well. It's pretty easy to infer from the eps file (although not as directly as the left one).

The bounding box info in the EPS file gives four integer values: X/Y coordinates of the lower left and upper right corners. 0/0 refers to the left lower corner of the staff symbol, so the negative first value is the actual left protrusion in points (1/72 inch, not the LaTeX `\pt`).

The third value is the horizontal position of the right edge of the image, and that is zero-based (as per the paragraph above. So since we know the line width we can determine a difference as "third value minus line-width in points", and if that is a positive value we have a (rather rare) case of protrusion.

One caveat that does *not* apply to the left margin: the points given in the EPS files are *integers*, not floats, which means there are rounding errors. Which in turn means there may be "false positives" that should be ignored

uliska:
Closes #36 
Closes #39 

Hopefully a comprehensive solution for horizontal alignment of scores.
I've added a description in the example file `examples/protrusion.tex` whose output is attached [here](https://github.com/jperon/lyluatex/files/1740447/protrusion.pdf). I'll base the manual section on that example.

Well, a few things are still open, but I thought they could be brought up either during review or even after merging this:

* I managed to produce a default of "marginwidth" that calculates the current right or left margin, so by default protruding elements won't be cut off at the page border.  
However, I would also like to be able to specify the values for the max-protrusion options as multiples of that margin, so the score may protrude e.g. "2/3 into the margin"
* The calculation of a margin width (both the full margin or a new "dimension") are done on the "current" page, so if a score crosses a page break the margins will be flipped, and dealing with that would be quite non-trivial: there may be constellations where the need for recompilation of a score (or a system) differs depending on whether it's on a left or right side).


rpspringuel:
> I managed to produce a default of "marginwidth" that calculates the current right or left margin, so by default protruding elements won't be cut off at the page border.
However, I would also like to be able to specify the values for the max-protrusion options as multiples of that margin, so the score may protrude e.g. "2/3 into the margin"

To produce this, save the calculated margin width(s) into a distance so that the user can use the normal LaTeX syntax for proportions of other distances.

It may then, however, be necessary to convert the stored value of the setting into a macro (as opposed to a glue or dimension register) which expands to look like a distance.  This is how we do things over in the Gregorio project.  All the various user settable distances are stored this way.  So long as they expand to the correct syntax for a glue or dimension, they can be used in most places where such things are expected (though judicious use of `\expandafter` or a temporary register might be necessary in some cases).  The reason we did this is that it allows the user to set a distance relative to some value which may change before we get to the score (such as the `em` or `ex` unit, or [even something more complicated](http://gregorio-project.github.io/tips/fontdimen.html)).  The same idea would work for values which aren't known until one reaches the score.

uliska:
I noticed that a traceback is produced for `insert=linline`. Obviously it tries to read the protrusion. I'll look into that.

uliska:
> I noticed that a traceback is produced for insert=inline. Obviously it tries to read the protrusion. I'll look into that.

Should be fixed in https://github.com/jperon/lyluatex/pull/142/commits/84f9690b634526db27c743036e98fe2e29405d57

uliska:
> To produce this, save the calculated margin width(s) into a distance so that the user can use the normal LaTeX syntax for proportions of other distances.

This has to be in the LaTeX side?

rpspringuel:
My LaTeX programming skills are better than my Lua programming skills, so I'm probably biased.

That said, I don't know that it *has* to be done on the LaTeX side, but that would be where I would normally think to put it.

At some point something like `0.75\leftmargin` (the usual LaTeX way of expressing proportions of a distance) is going to have to be converted to a distance.  If `\leftmargin` is defined on the LaTeX side, then the LuaLaTeX engine can do that conversion in a fairly straight forward manner using its existing capabilities.  If it isn't, then you're going to have to do the interpretation on the Lua side and swap out the `\leftmargin` (which is not a Lua expression) for `margin_left()` (which is) in order for the expression to make sense.  Further, you'll have to protect `\leftmargin` from the LuaLaTeX engine to prevent it from trying to expand an undefined macro before it passes the `max-left-protrusion` option to the Lua parser.

If you really want to do things on the Lua side, it might be easier to create a special syntax for this situation.  For instance, if `max-left-protrusion` is set to be a number without units, then the number is interpreted as a proportion of `margin_left()`.  See my comment in code.

uliska:
> If you really want to do things on the Lua side

No, I don't have any preference about this, it's just the other way round: I'm relieved when I'm able to leave TeX and write Lua (although I hadn't written a single line before this project ...)

The `'0'` approach is copied from `staffsize`

I thought of creating two lenghts so they can be used in the option and "catch" them in the `convert_unit()` function, but somehow this seemed strange and non-straightforward to me.
Another option would be to use `0.7leftmargin` and handle it analogue to the TeX units (default would then be `1leftmargin`. The two names would have to be registered somewhere and also be handled specifically in `convert_unit()`. Maybe this is the way to go.

jperon:
IMHO, the `.75\leftmargin` would sound "more familiar" to ordinary LaTeX users, but I've no strong opinion about this.

Another point is that, either in LaTeX or in Lua syntax, we should prefix those names: in LaTex, `\lyleftmargin` would be an option; in Lua, it should be a member of `ly` table.

> The calculation of a margin width (both the full margin or a new "dimension") are done on the "current" page, so if a score crosses a page break the margins will be flipped, and dealing with that would be quite non-trivial: there may be constellations where the need for recompilation of a score (or a system) differs depending on whether it's on a left or right side).

I think we could have two solutions to that problem:
- either take in account only the smallest margin, taking it as a maximum for both sides;
- or implement an approach similar to [solutions found here](https://tex.stackexchange.com/questions/6143/if-then-else-for-odd-page-even-page).

jperon:
Just pushed some small corrections. There's one important though: the line `self.ly_code = self:header()..self:content()` must stay as is for several reasons:
- `self.ly_code` is used in several places;
- without it, we have multiple calls to functions `Score:header()` and `Score:content()` without necessity.

uliska:
Am 20. Februar 2018 22:22:54 MEZ schrieb jperon <notifications@github.com>:
>IMHO, the `.75\leftmargin` would sound "more familiar" to ordinary
>LaTeX users, but I've no strong opinion about this.
>

When starting to write some code J came to try \lyleftmargin (\leftmargin is already taken).

I had to stop but I think it will work that way. 

>Another point is that, either in LaTeX or in Lua syntax, we should
>prefix those names: in LaTex, `\lyleftmargin` would be an option; in
>Lua, it should be a member of `ly` table.
>
>> The calculation of a margin width (both the full margin or a new
>"dimension") are done on the "current" page, so if a score crosses a
>page break the margins will be flipped, and dealing with that would be
>quite non-trivial: there may be constellations where the need for
>recompilation of a score (or a system) differs depending on whether
>it's on a left or right side).
>
>I think we could have two solutions to that problem:
>- either take in account only the smallest margin, taking it as a
>maximum for both sides;

I would be glad if that's not necessary.

>- or implement an approach similar to [solutions found
>here](https://tex.stackexchange.com/questions/6143/if-then-else-for-odd-page-even-page).

I will have a closer look at that question. But it won't fully help us with the problem that the page may change *during* a score.
This may cause problems when inner and outer margins are very different. The issue is that we don't only have to deal with a reduced overall line length but that the shortening depends on *two* aspects on the right side: not only the *overall* extent matters but also the staff must not protrude over the type area at all.

-- 

jperon:
> But it won't fully help us with the problem that the page may change *during* a score.

As I think the size of odd/even margins shouldn't change during the score (just their side), we could use something like `\ifodd` in `texoutput` variable of `latex.includesystems`. Just an idea.

uliska:
Am 20. Februar 2018 22:51:11 MEZ schrieb jperon <notifications@github.com>:
>Just pushed some small corrections. There's one important though: the
>line `self.ly_code = self:header()..self:content()` must stay as is for
>several reasons:

I noticed that too, and I will look into it, but I also made the change for specific reasons because of the recompilation issue.

>- `self.ly_code` is used in several places;
>- without it, we have multiple calls to functions `Score:header()` and
>`Score:content()` without necessity.

-- 

uliska:
>> [@jperon] Just pushed some small corrections. There's one important though: the
line `self.ly_code = self:header()..self:content()` must stay as is for
several reasons:

> [@uliska] I noticed that too, and I will look into it, but I also made the change for specific reasons because of the recompilation issue.

As I suspected when reading the code this "correction" doesn't work. If you compile this minimal example:

```TeX
\documentclass{scrartcl}

\usepackage{lyluatex}
\begin{document}

\begin{lilypond}[nofragment,
max-left-protrusion=1cm,
max-right-protrusion=1cm,
print-only=1%
]
{
  \set Staff.instrumentName = "Violin 1"
  \shape #'((0 . 0)(0 . 0)(3 . 0)(20 . 0)) Tie
  c'1 ~ \break c'
}
\end{lilypond}
\end{document}
```
you'll notice that it will be in an endless loop recompiling the score over and over again.

I'll explain why I did that change in the first place, and after that I think you'll either agree with me or can make a better suggestion.

When it is detected that we have to recompile a shortened score we *do* need to call `Score:header()` again to update the line-width (which your correction prevents, with the result that the score is compiled identically as before). But when this is done on `self.ly_code` the header is actually duplicated: upon recompilation the (new) header is prepended to the existing `ly_code` which already contains the (old) header. This is why I decided not to use `self.ly_code` in this processing loop but pass it directly to `run_lilypond()` as a local variable. In `run_lilypond()` the ly_code is only touched once so it doesn't have any negative impact.
Actually (if I'm not mistaken) when we're at the `if do_compile` in the `Score:process()` function `self.ly_code` will never be used again, so I don't see a reason why it shouldn't be used as a local variable/argument.

jperon:
So here is the real correction: without it, `\lilypondfile` failed with option `debug`.

uliska:
> So here is the real correction: without it, \lilypondfile failed with option debug.

Thanks, this seems to have been the fix.

uliska:
New observation: intermediate files are only deleted for the file version actually being used in the document. (I'll look into it)

uliska:
Next steps to be done are
* improve determination of odd/even
* perform an additional check in `includesystems` and
  at least adjust the horizontal offset (I think we can still find out the length of the systems and if a recompilation would *have been* necessary, so we can print a warning)

uliska:
@jperon @rpspringuel I tried to modify `latex.includesystems()` to a) tex.print each  systems directly rather than collecting the whole input and b) repeatedly checking for odd/even. This didn't work (https://github.com/jperon/lyluatex/commit/c45caba266f44d28aa5bf80d013c33c05c8dde59), and actually I'd wondered if it were even possible.

I know that the `ly.is_odd_page()` function is still too naive, but in my simple examples it still worked correctly.
Do you have an idea how it would be possible to detect changes of the current page *while outputting the systems*?

uliska:
Oops, I realize that my commit didn't even *try* to determine the *current* page - `ly.PAGE` is only set once at the beginning of the process.

But that gets us back to the question of how/whether it is possible to determine the current page *from within Lua*.

jperon:
> that gets us back to the question of how/whether it is possible to determine the current page *from within Lua*.

As far as I know (after trying for a while), it isn't; be would it be, it would give you the current page *at the beginning of Lua processing*. Indeed, the LaTeX code produced thanks to `tex.sprint` will be evaluated once Lua has finished its processing: within the same `\directlua{}` call, `ly.is_odd_page()` will always return the same result.

So if you want to evaluate page parity at each system, you'll have to define (on LaTeX side) a macro that returns the result you intend, which macro will be "tex.sprinted" from Lua. If this macro needs Lua code itself (which is possible), you'll get the following process:
LaTeX -> Lua -> LaTeX -> Lua.

I hope I was clear enough (it's not evident for me to explain in English…); in practice, I think you'll have to replace `\hspace*{=spt}` (l. 397) by a macro which will set the good space depending on page parity.

uliska:
> As far as I know (after trying for a while), it isn't [possible]; 

I've learned today that we can easily access the page counter from Lua through `tex.count['c@page']`. But ...

> be would it be, it would give you the current page at the beginning of Lua processing. Indeed, the LaTeX code produced thanks to tex.sprint will be evaluated once Lua has finished its processing: within the same \directlua{} call, ly.is_odd_page() will always return the same result.

... this I've learned too today. I [asked on stackexchange](https://tex.stackexchange.com/questions/416486/how-can-i-retrieve-thepage-or-detect-a-page-change-from-lua-in-luatex) and was pointed to that issue.
It seems we have three possible approaches *if* we want to tackle the challenge (more on the "if" below):

* Pass control back and forth between Lua and TeX like you described it. There is a way to accomplish this which I haven't fully understood on first reading:  https://tex.stackexchange.com/questions/20893/concurrently-interleaving-execution-of-lua-and-tex-in-luatex
The problem I see with this is that it's a complex operation that will require fundamental changes to the structure of our code - and it is not even guaranteed to work since the page counter is no reliable reference anyway.
* Basically move the code of `includesystems` to LaTeX: have the systems be prepared by Lua (as we have it) and then loop over them in LaTeX.
* Implement a label-ref approach that in a first run only sets labels and later (in a second run or somehow at the end of the compilation) determines page changes through their \pageref-s.
The elegance in this solution is also that it will work reliably when the scores are wrapped in floating environments.
Maybe (I hope) it is possible to access the pagerefs *during* the first LaTeX run and reliably determine if the scores have to be recompiled (once more).

I'm quite sure that the third option is the way to go if/when we want to fix this issue.
But the question is: what do we want and need to achieve *now*?

As long as there are fixed values for `max-left-protrusion` and `max-right-protrusion` (i.e. no notion of inner/outer) there is no problem, the alignment will correctly work on all pages and also over page breaks.
But as soon as we have inner and outer max-protrusion values we open a can of worms because the alginment *may* break when the sides flip after a page break.
And when we head towards max-protrusion values that are *relative* to outer or inner margins that can of worms gets huge because it can only be really solved by recompiling the score in parts while checking how the changed line breaking may actually change the distribution of the *content* over the systems.

My suggestion is that for the upcoming release we limit the functionality to `max-left-protrusion` and `max-right-protrusion` with fixed values (in TeX units). As a default value I would give up the idea of automatically determining the left/right margin (as that will already cause troubles with flipped pages) and set it to `\maxdimen` instead, effectively switching it off. This may cause score elements to be cut off at page borders - but we can provide a warning in that case, and this is what LilyPond itself does too.
All further functionality should be discarded for now and carefully described in separate feature requests for future releases.

If you agree @jperon I can "roll back" the code to such a state tomorrow morning.

uliska:
> My suggestion is 

Having slept over it I've come to the conclusion that this is the only way for us at this point. I'll do that and update the Pull Request this morning.

jperon:
> As a default value I would give up the idea of automatically determining the left/right margin (as that will already cause troubles with flipped pages) and set it to `\maxdimen` instead

Just an idea again, but taking the smallest margin instead of `\maxdimen` as a default would be a sensible default:

- I don't think most users would like a score to protrude more than this size (even in the bigger margin);
- it would avoid certainly unacceptable results (thing that get out of the page);
- as it would only be a default, those who *really* aren't satisfied with that would have the option to change it.

But do what you prefer !

uliska:
I had also thought about this, and your points are valid. Actually I think both ways would be "correct". What made me think of "my" solution is

* it's much simpler
* it's the behaviour people are used to from LilyPond
* it's consistent with `insert=fullpage` (where you don't have *any* option to achieve what we have with `max-protrusion` anyway)

I'll prepare the PR with the simple implementation, and we can *then* see whether to still update it.

uliska:
I hope this is now good to go, with the limited functionality discussed above (note that IISC "limited" is still more than any other solution provides). The behaviour is documented on the `update-manual` branch, for reference I've attached an [excerpt from the compiled manual](https://github.com/jperon/lyluatex/files/1747183/protrusion-doc.pdf)

jperon:
When compiling test.tex, there is a problem with margins with `insert=systems` from page 7: margins get inverted.

[test.pdf](https://github.com/jperon/lyluatex/files/1747239/test.pdf)

uliska:
Should be fixed now. I'm pretty sure this is an old bug that is unrelated to protrusion since both the left and the inner margin of the LilyPond document had been set to TeX's "inner" margin.

PR #148, opened 2018-02-21
Suppress "overfull hbox" warnings

uliska:
We will very often produce overfull hboxes through intended protrusion.
The way to completely suppress the warnings is to temporarily set `\hfuzz` to
a huge value.

Two questions about the approach:
* is it OK to "hijack" `\ly@setunits` for this or should it be implemented in its own macro?
* Is it OK to *completely* suppress the warning?
  I think so because we shorten a score not to be cut off at page border and if the user sets max-protrusion to be even higher they should be knowing what they do.


jperon:
OK for me. If we want to be scrupulous, we could have our own warning in case the material gets out of page.

uliska:
I'll add a warning when max-protrusion has been exceeded, but I'll do so on the`ptrotrusion` branch.

Issue #139, opened 2018-02-17
Empty options still requires brackets


rpspringuel:
If the default options are satisfactory, the user should not be required to put `[]` in the code as this is contrary to general LaTeX practice.  We may be stuck with this for the moment, but it is something that should be investigated and fixed at some point down the road.

jperon:
I totally agree with that. Here is a mwe (!) so that we can investigate: if you replace `\newenvironment{ly}[1][]{%` by `\newenvironment{ly}{%`, it gives the expected result: no pdf output, but two lines on stdout. Perhaps it's a bit intrepid from me, but I'm wondering whether this shouldn't be considered as a bug in LuaTeX's `process_input_buffer` callback…

```TeX
\documentclass{minimal}
\usepackage{luacode}

\begin{document}

\begin{luacode*}
ly = {}
ly.score_content = {}
function ly.env_begin(envs)
    function ly.process_buffer(line)
        table.insert(ly.score_content, line)
        for env in envs:gmatch('[^,]*') do
            if line:find([[\end{]]..env:gsub('^ ', '')..[[}]]) then return nil end
        end
        return ''
    end
    ly.score_content = {}
    luatexbase.add_to_callback(
        'process_input_buffer', ly.process_buffer, 'readline'
    )
end

function ly.env_end()
    luatexbase.remove_from_callback('process_input_buffer', 'readline')
    table.remove(ly.score_content)
end
\end{luacode*}

\newcommand\lyscorebegin[1]{\directlua{ly.env_begin('#1')}}
\newcommand\lyscoreend{\directlua{ly.env_end()}}

\makeatletter
\newenvironment{ly}[1][]{% %% THIS LINE SEEMS TO CAUSE THE PROBLEM
  \lyscorebegin{\@currenvir}
}{%
  \lyscoreend
  \directlua{
    print('\string\n\string\nCONTENT:')
    print(table.concat(ly.score_content, '\string\n'), '\string\n')
  }
}
\makeatother

\begin{ly}
  a b c
  d e f
\end{ly}


\end{document}
```

PR #153, opened 2018-02-22
Refactor

jperon:


uliska:
@jperon #156 should be fixed before merging this PR.

test.tex
```diff
@@ -19,13 +19,14 @@
 
 % It's possible to wrap `ly` within another environment, using \ly[] and \endly.
```
rpspringuel: Comment no longer matches content.

jperon:
e56a803 didn't work when there where several options with braces, for example:

```TeX
\lilypondfile[includepaths={./,test}, print-only={3, 1, 5}]{ly/Ave Maria De Lourdes/Ave Maria De Lourdes}
```

So I pushed another fix.

jperon:
@uliska Could you please test it ? If it works, I think we can merge.

jperon:
In fact, I've just seen that #139 isn't really fixed; so please don't merge at once.

uliska:
#156 can be considered fixed, but I won't merge as per your last comment.

jperon:
Now that should be good. #139 isn't perfectly fixed, but I think it won't be totally fixable as is. The reason is that LuaTeX's `process_input_buffer` needs a newline to begin processing lines; but LaTeX deletes the newline after `\begin{env}` when `env` has only optional arguments and no `[]` is specified.

PR #154, opened 2018-02-23
Raw pdf name output

uliska:
This option does not return any \includegraphics macros or similar but the raw PDF name(s) (without extension).

For `insert=inline|bare-inline|fullpage` the plain filename is returned, including the tmpdir and the hash (i.e.: `self.output` of the score). 
For `insert=systems` a comma-separated list is returned, containing the file names as above but also the horizontal offset in the first element:
```
-5pt,tmp-ly/my-hash-1,tmp-ly/my-hash-5
```
(when `print-only` is set to `{1,5}` and the left protrusion is 5pt).

Open questions:

* I do currently *not* enclose the return value in braces because I think it is better to consume the plain filename string
* At this point my LaTeX fu is not sufficient to create appropriate wrapper commands. I don't know how I can create a command or environment where I can enter the lilypond code, pass it to lyluatex functions and "use" the result (e.g. by wrapping it in my own \includegraphics calls). All I managed was to print the raw output to the document (proving that the code itself works properly). It would be great if someone could add such examples to `examples/wrappingcommands.tex`

**PS:** This is based on #153 and can only be merged after that.

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -386,6 +386,12 @@ function latex.label(label, labelprefix)
     if label then tex.sprint('\\label{'..labelprefix..label..'}%%') end
 end
 
+function latex.systems_list(filename, hoffset, range)
+    tex.sprint(-hoffset..'pt,')
+    for i = 1, #range - 1 do tex.sprint(filename..'-'..range[i]..',') end
+    tex.sprint(filename..'-'..range[#range])
+end
+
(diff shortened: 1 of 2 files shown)


uliska:
I've now merged the new master branch into this, and I think it should be good to go now. Although I'd still need a proper example file and help how to make use of the raw filenames in other packages.

PR #155, opened 2018-02-23
Better handling of suppressed indentation

uliska:
calc range already in the compilation loop
- check for indent there (not when writing systems to LaTeX)
- deactivate indent when
  - only system no. 1 is printed
    (because score only has one system or due to print-only)
  - the first system is *not* printed first (through print-only)
- if indent is deactivated and the score has more than one system
  then recompile without indent to have a properly filled system
  (and not one that is just shifted left)

See new example file

Based on #153

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,10 @@ sympy-plots-for-*.tex/
 # WinEdt
 *.bak
 *.sav
+
+# LilyPond intermediate files
+*-systems.*
+*.midi
+
+# lyluatex temp directories
+*tmp-ly/
--- a/examples/dynamic-indent.tex
+++ b/examples/dynamic-indent.tex
@@ -89,7 +89,26 @@ will be automatically activated to avoid having the \emph{whole} score narrowed:
   \repeat unfold 25 { c' d' e' d' }
 \end{lilypond}
 
+\paragraph{Right protrusion}
+The dynamic handling of (automatic) indent also works correctly when there is
+protrusion handling to the right. The following score has the ties manually
+shaped to exceed the staff symbol by 10, and 7 staff spaces, and
+\texttt{max-protrusion=1cm} .
 
+\begin{lilypond}[nofragment,max-protrusion=1cm,]
+{
+  \set Staff.instrumentName = "Violin 1 & 2"
+  \set Staff.shortInstrumentName = "Vl 1 & 2"
+  \shape #'((0 . 0)(0 . 0)(3 . 0)(10 . 0)) Tie
+  c'1 ~ \break
+  \shape #'((0 . 0)(0 . 0)(3 . 0)(7 . 0)) Tie
+  c' ~ \break
+  c'
+}
+\end{lilypond}
+
+
+\paragraph{Performance considerations}
 The handling of indent suppression may require up to four compilations of the
 score, but these are handled automatically, and the resulting intermediate
 stages of the score are cached just like the scores actually used in the
--- a/examples/dynamic-indent.tex
+++ b/examples/dynamic-indent.tex
@@ -101,5 +101,16 @@ more LilyPond compilations and therefore compilation time. But it will only
 apply and be executed if the score exceeds the protrusion limit, so it can only
 occur in circumstances where multiple LilyPond runs are expected anyway.
 
+\paragraph{Known Issues:}
+If \texttt{autoindent} is implicitly set like in the following example, and
+\texttt{max-protrusion=0cm} then the indented first system doesn't correctly
+refer to the text margin (\texttt{=0cm}) but to the staff symbol of the
+surrounding systems.
+
+\begin{lilypond}[indent=1cm,print-only={3,1,2},max-protrusion=0cm]
+  \set Staff.instrumentName = "Violin"
+  \repeat unfold 25 { c' d' e' d' }
+\end{lilypond}
+
 
 \end{document}
--- a/examples/dynamic-indent.tex
+++ b/examples/dynamic-indent.tex
@@ -8,26 +8,29 @@
 This document demonstrates the use of \texttt{indent} and \texttt{autoindent},
 partially in combination with \texttt{print-only}.
 
-\texttt{indent=1cm} indents the first line, but if the resulting score contains
+\texttt{indent=2cm} indents the first line, but if the resulting score contains
 only one system this indent is suppressed (issuing a warning on the console):
 
-\begin{lilypond}[indent=1cm]
+\begin{lilypond}[indent=2cm]
+  \set Staff.instrumentName = "Violin"
   \repeat unfold 12 { c' d' e' d' }
 \end{lilypond}
 
-\begin{lilypond}[indent=1cm]
-  { c' d' e' d' }
+\begin{lilypond}[indent=2cm]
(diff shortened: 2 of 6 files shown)


uliska:
I notice that sometimes tracebacks occur related to comparing a number to `protrusion` (a string value). This happens when a document is compiled but *some* things are changed.

I'm not sure if that can happen when only the user has changed things or if it only occurs when I did change the package code. Anyway, I'll have to look into this later tonight.

uliska:
I've fixed the traceback issue, but I'm currently trying to make it even smarter. So this isn't ready to be merged ATM.

uliska:
This should now be good to go. It was surprisingly challenging to implement my idea - especially considering that it is only useful for really rare corner cases. Probably it would have been more straightforward to simply issue a warning telling the user to remove the `indent` option for that score ;-)

However, it was fascinating that it *is* possible to achieve something like that. See the compiled file 
[print-only.pdf](https://github.com/jperon/lyluatex/files/1755235/print-only.pdf) for reference

jperon:
There where some problems, that should be fixed with last commit.
@uliska Could you please check that I didn't break anything you had done ? If not, I think it can be merged.

uliska:
No, you don't seem to have broken anything (although I don't fully understand all your changes).
And no, it can't be merged yet: I realized that having protrusion limit on *both* sides can enter a compilation loop (and this already was the case before your corrections). So I'll have to look into it once more ...

jperon:
Could you please give me a mwe ? I've already suppressed some infinite loops, so perhaps I'll get it.

jperon:
Just a question: `max-left-protrusion` and `max-right-protrusion` are in `HASHIGNORE`; couldn't it lead to unwanted results (for example if someone changes those to 0 after first compilation) ?

uliska:
> Could you please give me a mwe ? I've already suppressed some infinite loops, so perhaps I'll get it.

Test with `examples/protrusion.tex`.

However, this file exposes an issue with your last state: Files *without* explicit indent entered the "deactivate indent" loop and caused a traceback in the recompilation. This had to be fixed in b318af67e41a7865562c70f131d774fcf2266db5. After that fix we have the loop, before that it was a crash.

This is a MWE:

```TeX
\documentclass{scrartcl}
\usepackage{lyluatex}
\begin{document}

\begin{lilypond}[nofragment,
print-only=1,%
max-protrusion=0.24cm]
{
  \set Staff.instrumentName = "Violin 1"
  \shape #'((0 . 0)(0 . 0)(3 . 0)(10 . 0)) Tie
  c'1 ~ \break c'
}
\end{lilypond}
\end{document}
```

Note that when you abort the loop and then recompile the whole document it runs through on first attempt (so it correctly picks up the cashed score).

uliska:
I think the cause for the loop is the fact that check_indent() gets 'shorten' as the argument. This was good for the original protrusion handling but not for the indent: here we need to consider the sides independently. So we need to get shorten_left and shorten_right as separate variables.

uliska:
Another thought: now that we've invested so much time into this topic I'm wondering about the use case.
What this PR implements is a corner case, but I've just noticed a use case that could be really useful and should be made available as an additional option: auto-indent.
I think it is a regular case that a protrusion overflow is triggered by the first system. I suggest a mode where in case of left overflow we don't shorten the line width but add indent. If after that there is too much protrusion left (from other systems) line width is shortened with indent adjustment.
Basically this is what the current PR is doing - but applied to all scores, also with the first system in the regular position.
I think this will often result in a much smoother appearance than the ordinary shortening.

jperon:
With the mwe, I don't exactly get an infinite loop: it compiles 10 times (which is much for one score !), but eventually succeeds; and when compiling again the mwe, it successfully uses the cache, so that the compilation is very fast.

uliska:
In that context I would suggest to drop the special handling of "system 1 is printed at a later position". This is really obscure, and users should simply disable indent in such cases.
That would simplify that complex conditional, but we could reuse most of the calculations in the new option too.

jperon:
Just done. The key line to fix the loop is:

`until self:check_protrusion(bbox.read) or self:check_protrusion(bbox.parse)`

The reason is that, after first compilation, the `.eps` are removed; so `bbox.parse` can't return expected values.

jperon:
I let the special handling of "system 1 is printed at a later position": with a clear comment, and as the calculations serve anyway for `autoindent`, I think it's worth keeping it.

jperon:
And a last word: what a great work you did!

uliska:
Thank you. I tracked down some issues when writing a proper example file. I'll look into these later today.

uliska:
I'll need some more time for this because it is really quite tricky.
[dynamic-indent-initial-bugs.pdf](https://github.com/jperon/lyluatex/files/1760219/dynamic-indent-initial-bugs.pdf) shows the compilation result and issues at the point of 7375e748507406016b0bbb5fc2c498d2cd5447d9
Please don't do anything about it because I have already done quite a lot of uncommitted changes, and I'm committed to bringing this to an end ;-)

uliska:
OK, I hope I've finally managed to get to grips with this :-)
I had to rebuild this from the ground up so please check carefully.  But I can say that

* all tests I did worked correctly (see attached [rendering](https://github.com/jperon/lyluatex/files/1763931/dynamic-indent.pdf) of `examples/dynamic-indent.tex`)
* Right protrusion is properly supported
* All examples seem to have taken exactly the expected number of recompilations
* All intermediate scores are cached and not recompiled

The most important change with regard to the interface is that I made `autocompile` default to `true`. I know it's quite “expensive”, *but*: autoindent will *only* apply and be executed when a `max-protrusion` is set and exceeded - so it will only use that addition time for recompilation when the user already has to expect (and therefore accepts) the possibility of recompilation.
Additionally I've removed the `noautoindent` option, but only because it made things more awkward and `[autoindent]` didn't set it to `true` (`[autoindent=true]` was required).

One minor issue remains to be fixed (it is described at the end of the example file): when `max-protrusion=0cm` and autoindent is activated implicitly after deactivating a regular indent the alignment is not fully correct: the indent will be chosen so that first system aligns with the remaining staff lines, not with the text margin.

uliska:
One thing I forgot to mention in the previous comment: `autoindent` is set to `true` by default, but when `indent` is set to any value `autoindent` will be deactivated.
However, if the indent is deactivated through `print-only={3,1,2}` and this causes a protrusion overflow then not the *whole* score will be shortened but `autoindent` activated implicitly.

jperon:
As stated in last commit, I just had to make a small change to avoid an error with `\includely`. Other changes are merely cosmetic ones.

If you don't have anything to add, I think we should merge it. As for the "issue", I wouldn't regret it if even we considered it as a wontfix, as it is really a corner case.

uliska:
> As stated in last commit, I just had to make a small change to avoid an error with \includely. 

Could you please explain that in some more detail, please?

> Other changes are merely cosmetic ones.

Yes, I've seen that.

> If you don't have anything to add, I think we should merge it. As for the "issue", I wouldn't regret it if even we considered it as a wontfix, as it is really a corner case.

I'd say we should add an issue stating the fact (and move the mention from the example document there) but don't have it block the merge or the release.

PR #158, opened 2018-02-25
#156 parse braced values with or without commas

### LINKED ISSUE #156: Regression on option parsing

uliska:
Closes #156

As commented in https://github.com/jperon/lyluatex/issues/156#issuecomment-368293867 I think it isn't possible to use a simple iterator regex to match the different types of option values.

I hope this implementation (although more complex) finally does the right thing.

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -1372,27 +1372,32 @@ end
 
 function ly.set_local_options(opts)
     local options = {}
-    local next_opt = opts:gmatch('([^,]+)')  -- iterator over options
-    local opt = next_opt()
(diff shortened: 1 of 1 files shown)


uliska:
#156 is fixed in https://github.com/jperon/lyluatex/commit/15cd4b933e4c14de6bcbf87f63f34af54747cb5f, therefore closing this PR

PR #162, opened 2018-02-28
Better handling of current-font-as-main

uliska:
While reviewing the manual section about fonts I changed some small details about the handling of `current-font-as-main`:

- default value is now 'false'
- if `rmfamily` is set then current-font-as-main is unconditionally set to false
- but only when it *is* set in the document a warning is issued

More explanations in https://github.com/jperon/lyluatex/commit/8b9bc1952eef1b94c5c09ca2a71f77099d9a2edf


uliska:
The commit is also included in `update-manual` branch, but I wanted to have it reviewed separately.

PR #173, opened 2018-03-04
Rounding issues

### LINKED ISSUE #170: Corner case problem with dynamic indent
OMG, I've just noticed a regression with `examples/dynamic-indent.tex`:

image

It seems in the second stage the indent isn't reduced appropriately.

uliska:
Closes #170

This PR includes two improvements to the handling of protrusion and indentation:

* Correctly convert the PostScript Units from the EPS bounding box to LaTeX points
* Retrieve the HiResBoundingBox from the PDF (if Ghostscript is available)

Both together make the example from #170 compile correctly with all `DIV` values from 6 through 20, so I consider the issue solved. But please test with random other examples if you notice any regressions.


lyluatex.lua
```diff
@@ -262,74 +273,40 @@ function bbox.get(filename, line_width)
...
-    x_1, y_1, x_2, y_2 = get_bb()
     local bb = {
-        ['protrusion'] = -x_1,
-        ['r_protrusion'] = x_2 - line_width,
-        ['width'] = x_2,
-        ['height'] = y_2 - y_1
+        ['protrusion'] = -convert_unit(x_1..'bp'),
+        ['r_protrusion'] = convert_unit(x_2..'bp') - line_width,
+        ['width'] = convert_unit(x_2..'bp'),
+        ['height'] = convert_unit(y_2..'bp') - convert_unit(y_1..'bp')
```
uliska: Are you sure it is appropriate to exchange the PostScript Unit conversions with `bp`? They are quite close but not exactly the same.

jperon: According to [this](https://raw.githubusercontent.com/tweh/tex-units/master/tex-units_preview.png), 1bp = 1/72in.

In fact, if I look at the result of `\directlua{print('\string\n', (tex.sp('1in')/tex.sp('1bp')) / 72)}`, I get `1.0000114014685`, which means an error of 0.0011%. On the width of an A4 sheet, this means an error of 0.0024mm.

PR #178, opened 2018-03-10
WIP: Make clean (First step to CTAN build)

uliska:
We have to pretty much review the ctan target in the Makefile.
What I think we should do is first clean the repository, then build
the manual (i.e. force a rebuild of everything to avoid
- getting outdated PDFs into the release
- missing errors because existing PDFs are used)

I open the Pull Request already now because I'm not very familiar with Makefiles.
The next step(s) will be to copy the appopriate files and directories to the .ctan directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,8 +11,13 @@ manual:
 		makeindex lyluatex && \
 		lualatex --shell-escape --interaction=nonstopmode lyluatex.tex
 
+clean:
+	rm -rf tmp-ly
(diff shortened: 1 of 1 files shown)


jperon:
I'm just working on that question (from this morning); please wait a little, as I already have made quite important changes.

jperon:
I'm just working on that question (from this morning); please wait a little, as I already have made quite important changes.

PR #179, opened 2018-03-10
Ctan

jperon:
This PR includes in CTAN only what is necessary, merging examples within lyluatex.tex (produced by pandoc), and not including lyluatex.md.


jperon:
I've forgotten to mention that with those changes, if you want to compile one example, you have to run `./compile-example.lua example` from the `examples` directory.

jperon:
With last commit, examples can be compiled with `latexmk`.

README.en.md
```diff
@@ -41,46 +41,46 @@ The argument `staffsize`, which is optional, changes the size of the score.  You
...
+    \lilypond[staffsize=12]{c' d' g'}
 
-**Nota bene:** The `\lily` command *does not* support blocks of LilyPond code
-with explicit `\score` blocks.  Such code must be included with the `ly`
-environment or as a separate file.
+**Nota bene:** The `\lilypond` command *does not* support blocks of LilyPond
+code with explicit `\score` blocks.  Such code must be included with the
+`lilypond` environment or as a separate file.
 
 See the document `test.en.tex` for an example.
```
uliska: Either this reference should be deleted or the file be copied to the ctan folder in `make ctan`.

jperon:
Just pushed the modifications.

uliska:
Sorry that I was so inactive over the weekend.

I'm basically fine with that PR. However, I'd like to ask you to reconsider one thing (note the difference to "request changes" please): I would prefer if the example files would remain standalone LaTeX files, without the `lyluatexexample` dependency. So instead of the `compile-example.lua` approach I'd prefer to see `latex-flatten.py` to strip everything around the `document` environment, which should be easy to do as well.
It could then be decided whether the example files should also be copied to the CTAN directory or if there could be a reference to the directory on Github in the manual.

The reason for this is that I had always intended the examples directory as a collection of "sandbox" files that the users could open, inspect and play around with. This is much less possible if they are silently merged into the manual's `.tex` file. This can also include further example files that don't make it into the manual itself.

What do you think, @jperon?

jperon:
I've just changed it: the examples are ordinary LaTeX docs again, and `insert-examples.lua` makes the hard work when compiling the manual.

I've added a link to the repository for examples: I think that it's better to send as few files as possible to CTAN.

uliska:
Thank you very much, I like this better.

I had to struggle a little bit to make it work because `#!/usr/bin/env lua` didn't find my Lua installation. I had to write a wrapper script in my `bin` directory, then it worked. Maybe we should add that information to a file in the repo (which doesn't get copied to the ctan folder).

PR #196, opened 2018-04-26
Add first-page-number and print-first-page-number options

uliska:
a)
There may be the need to force a first page number (i.e. the start
with an odd or even page), especially in combination with the raw-pdf
option, when the resulting score is not immediately included in a score.
I came across this in a case where I want to first generate a number of
scores and afterwards include them in a row of documents (=> a set
of all parts for a score). I know that the scores will all start on an odd page
(because I'll take care of that) so I want to force LilyPond to think so,
regardless of what lyluatex thinks the "current page" is.

b)
Incidentally I think there may be reasons where the first page number may
*not* be wanted, even when included in a book.

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -994,15 +994,20 @@ end
 function Score:ly_paper()
     local papersize = '#(set-paper-size "'..(self.papersize or 'lyluatexfmt')..'")'
     if self.insert == 'fullpage' then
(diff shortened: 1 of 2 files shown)


uliska:
BTW, @jperon do you think there would be a way to *directly* call `lyluatex`'s functionality from some other Lua package, rather than using `tex.print()` to write lyluatex commands back to the document?

PR #201, opened 2018-05-30
Document local installation of current version

uliska:
As per http://tug.org/pipermail/tex-live/2018-April/041493.html the description of the local installation wasn't fully correct.

Actually it's a trivial change, but I'm still not fully sure I haven't misunderstood some details - therefore a pull request.

### DIFF
--- a/lyluatex.md
+++ b/lyluatex.md
@@ -55,7 +55,7 @@ tlmgr install lyluatex
 Copy `lyluatex.sty` and `lyluatex.lua` from this repository into your
 `TEXMFHOME` tree, or clone this repostory into your `TEXMFHOME` tree using Git.
 In many cases this will be `$HOME/texmf`, and \lyluatex\ should be located below
-`$TEXMFHOME/tex/luatex`. It is important that this is the `luatex` tree because
+`$TEXMFHOME/tex/luatex`. It is important that this is the `luatex` tree:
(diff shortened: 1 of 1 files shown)


lyluatex.md
```diff
@@ -52,10 +52,12 @@ tlmgr install lyluatex
...
 
-Copy `lyluatex.sty` and `lyluatex.lua` from this repository into your `TEXMF`
-tree, or clone this repostory into your `TEXMF` tree using Git, then run
-`mktexlsr`. Note that in this case your local copy will shadow the version
-possibly installed in your \TeX\ distribution.
+Copy `lyluatex.sty` and `lyluatex.lua` from this repository into your
+`TEXMFHOME` tree, or clone this repostory into your `TEXMFHOME` tree using Git.
+In many cases this will be `$HOME/texmf`, and \lyluatex\ should be located below
+`$TEXMFHOME/tex/luatex`. It is important that this is the `luatex` tree because
+otherwise \LuaLaTeX\ will still find the version from TeXLive first and use that
```
rpspringuel: You only need one of `because` and `otherwise`.  `otherwise` is more appropriate here because you give the counter example (what will go wrong if you don't do it right).  `because` would be for actually providing the reason why to do it the right way (i.e., engine specific distro trees have higher priority than generic local ones).

PR #216, opened 2019-02-07
Fix #202

### LINKED ISSUE #202: Relative inlucdes
I just tried recompiling an old project which made use of `#(ly:set-option 'relative-includes #t)` inside the LilyPond files and its unable to find the included files.  Looking through the log I see the following directory in the search path: `./Users/ss/Documents/Abbey/DivineOffice/Focus 2018/6-Friday/3-Vespers/music/hymns/`  This is an incorrect amalgamation of the directory where the tex file is located (`./Users/ss/Documents/Abbey/DivineOffice/Focus 2018/6-Friday/3-Vespers/`) and the one where the ly file is located (`./Users/ss/Documents/Abbey/DivineOffice/Focus 2018/music/hymns/`).  Since the tex file location appears 3 other times in the list and the ly file location doesn't (which is the whole problem), I suspect that it's the ly file location that's being mangled at some point.

jperon:
This should fix the problem with relative includes when the score isn't under the current dir.
Sorry, I had made mistakes before.

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -897,7 +897,7 @@ function Score:lilypond_cmd(ly_code)
         "-djob-count=2 "..
         "-dno-delete-intermediate-files "
     if self.input_file then
-        cmd = cmd..'-I "'..dirname(self.input_file):gsub('%./', lfs.currentdir()..'/')..'" '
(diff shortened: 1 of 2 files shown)


jperon:
In fact, this change is so trivial that I'm going to merge now.

PR #225, opened 2019-04-09
#224: Let LISTFILE respect tmpdir option

### LINKED ISSUE #224: FILELIST doesn't respect tmpdir option
The `.list` file with the list of score files is always saved in a `tmp-ly` directory below the current working directory, regardless of the `tmpdir` option.

AFAICS this is because the `FILELIST` variable is populated in `ly.declare_package_options`, i.e. *before* the options are processed.

uliska:
LISTFILE was set to the *default* value of the tmpdir option because
writing \ExecuteOptionsX to the tex document doesn't update the option.
This has to take place in a new \directlua call.

Fixes #224

### DIFF
--- a/lyluatex.sty
+++ b/lyluatex.sty
@@ -123,7 +123,7 @@
 \directlua{ly.make_list_file()}
 \directlua{
   if ly.get_option('cleantmp') then
-    luatexbase.add_to_callback('stop_run', ly.ly.declare_package_options, 'lyluatex cleantmp')
+    luatexbase.add_to_callback('stop_run', ly.clean_tmp_dir, 'lyluatex cleantmp')
     luatexbase.add_to_callback('stop_run', ly.conclusion_text, 'lyluatex conclusion')
   end
 }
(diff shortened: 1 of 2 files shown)


lyluatex.sty
```diff
@@ -120,9 +120,10 @@
     ['xml2ly'] = {'musicxml2ly'},
   })
 }
+\directlua{ly.make_list_file()}
 \directlua{
   if ly.get_option('cleantmp') then
-    luatexbase.add_to_callback('stop_run', ly.clean_tmp_dir, 'lyluatex cleantmp')
+    luatexbase.add_to_callback('stop_run', ly.ly.declare_package_options, 'lyluatex cleantmp')
```
jperon: Why do you replace this callback ? At the end of compilation, there's no need to process options, while it's necessary do clean the tmp dir.
```suggestion
    luatexbase.add_to_callback('stop_run', ly.clean_tmp_dir, 'lyluatex cleantmp')
```

uliska: Hm, I have no idea how that happened. Certainly not intentionally. Sorry

PR #229, opened 2019-04-13
Fix #223

### LINKED ISSUE #192: "Crash" with large (high/long) scores
I got an issue with a large score I wanted to compile with lyluatex and that "crashed" with the following error:

```
lyluatex.lua:126: Dimension too large
stack traceback:
	[C]: in function 'sp'
	...e/uliska/texmf/tex/latex/latex-git/lyluatex/lyluatex.lua:126: in function '
convert_unit'
	...e/uliska/texmf/tex/latex/latex-git/lyluatex/lyluatex.lua:325: in function '
parse'
	...e/uliska/texmf/tex/latex/latex-git/lyluatex/lyluatex.lua:287: in function '
bbox_func'
	...e/uliska/texmf/tex/latex/latex-git/lyluatex/lyluatex.lua:758: in function '
check_protrusion'
	...e/uliska/texmf/tex/latex/latex-git/lyluatex/lyluatex.lua:1110: in function 
'process'
```

Some `print()` debugging revealed the following values in the bounding box:

```
x_1    -35.19
x_2    435.929329
y_1    -21694.802
y_2    -0.57266199999867
```
which led me to have a look at the generated PDF (actually the score had been processed but not included in the document). And I realized that in the (unindexed) PDF the whole score is printed on *one* page, resulting in a file with dimensions 165x3017mm, which makes me think that the `y_1` value exceeds some TeX limits.

What should we do about this? Without digging into the code again (yet) I have the impression that we will never need to actually have this value, the total height of all systems glued together. Well, do we actuallly need the *height* at all in any place?
I see two possibilities:
a) don't use the unindexed pdf but the file `-1.pdf` (I think this code is only run if we already know there *is* a system)
b) don't read the *height* values at all.

What do you think, @jperon?

### LINKED ISSUE #223: \lilypond{} command broken?
I tried to compile the manual after pulling the latest master (at 1b94b9d0b95d5357e04e5c01b0020a8248fdd1af).
The compilation stopped with the following error:

```
(lyluatex)	Compiling score tmp-ly/c39a38c257f953808c7606e3e7c1f1b8 with LilyPond executable 'lilypond'.
./lyluatex.lua:366: attempt to perform arithmetic on local 'height' (a nil valu
e)
stack traceback:
	./lyluatex.lua:366: in function 'includeinline'
	./lyluatex.lua:1239: in function 'write_latex'
	./lyluatex.lua:1155: in function 'process'
                                                  
l.321 \lilypond{ c' d' e' }
                          Fragments specified with \cmd{lilypond} are by
?
```

The error can be reproduced with this MWE:

```tex
\documentclass{article}
\usepackage{lyluatex}
\begin{document}
\lilypond{c'}
\end{document}
```

I didn't notice anything with the `lilypond` *environment*, and I didn't do any substantial tests since I have just begun trying to incorporate `lyluatex` again in a project.

jperon:
I had to change the way bbox is stored on disk for future compilations.

Incidentally, better fix #192.

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -17,7 +17,6 @@ local ly = {
     err = err,
     varwidth_available = kpse.find_file('varwidth.sty')
 }
-local obj = {}
 local Score = {}
 
 local FILELIST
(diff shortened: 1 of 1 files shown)


jperon:
@uliska Did you launch `make clean` before recompiling ? I had this error once, but it disappeared after that.

PR #232, opened 2019-04-13
Customization commands

jperon:
This PR includes several tools for customization:

- `\lynewenvironment` lets one define wrappers for `ly` without most of the drawbacks mentionned in the doc (especially the `[]` for empty optional arguments);
- as `\lynewenvironment` doesn't allow adding lilypond code between `\begin{ly}` and `\end{ly}` (but only latex code before and after), four new parameters allow it in various places, which must be defined to a filename (or a list of filenames) to include (without '.ly'):
  - `include_header`;
  - `include_footer`;
  - `include_before_body`;
  - `include_after_body`;
- `\lysavefrag{FILE}` allows defining lilypond code to be saved to `tmp-ly/FILE.ly`, so that it may be called from `include_...`.

### DIFF
--- a/examples/wrappingcommands.tex
+++ b/examples/wrappingcommands.tex
@@ -2,7 +2,11 @@
 
 \begin{document}
 
+\VerbatimFootnotes
+
 \section*{Wrapping Commands}
+
+\subsection*{Command within commands}
 
 \cmd{lily} can be wrapped within another command as usual:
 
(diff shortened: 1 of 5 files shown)


jperon:
Here is a MWE:

```latex
\documentclass{scrartcl}
\usepackage[debug]{lyluatex}

\begin{document}

\lynewenvironment{myly}[1][]{%
	{\centering test \par}
	\begin{ly}[
		include_before_body={foot,foot},
		include_after_body=head,
	]
}{
	\end{ly}
}

\begin{lysavefrag}{head}
a b c
\end{lysavefrag}

\begin{lysavefrag}{foot}
g a' b
\end{lysavefrag}

\begin{myly}
d e f
\end{myly}

\end{document}
```

uliska:
This looks very interesting. I'll look into it, but most probably not today.

uliska:
I admit I don't really understand the code changes, so I can only state that both the MWE and the manual work with the modified content.

I have applied a few clarifications to the manual because I think people will be confused by a `caption` argument with the actual content "caption", or with "saved fragments" that are all named like they are keywords.

So from my perspective this can can be merged.

PR #233, opened 2019-04-15
Fix #230

### LINKED ISSUE #230: Separate out paper settings?
I've started working with `lyluatex` again in a large-scale project, and I must say I'm amazed how (some glitches admitted) easy and powerful the integration actually works. (I think there will be some further enhancements I'll contribute over the coming months).

One thing I notice is: I'm basically working with `\lilypondfile` in order to maintain my music examples separately in the directory tree and to be able to prepare them using Frescobaldi. What I'm looking for is a way to automate the process of having the same layout parameters in Frescobaldi as when compiling through `lyluatex`.

I have some ideas and could probably get something done for my project at hand, but ideally there would be  a generally usable solution to be included in `lyluatex`. The basic idea would be to have one resource that is used from both perspectives.

The approach that currently looks the most promising to me (although it would probably need quite some validation and fallback procedures) is (only for `\lilypondfile`):

* Do *not* write the generated paper variables to the generated .ly file
* Write it to an include file instead and place this next to the requested LilyPond input file
* Write an `\include` statement pointing to the previously generated file into the temp file

This way the actual paper variables are available from the original input file after the first run of `lyluatex`. 

Issues I can see right now:

* The include file has to be included by the main input file.
    * The user would have to actively do that (error prone)
    * It has to be ensured that this doesn't lead to *two* `include` statements in the .ly file generated by `lyluatex`
* This approach would probably fail when one example is to be used from multiple documents
* It might be considered intrusive to create a permanent file in the user file's directory. (And we have to consider possible conflicts with existing files). So it should be optional and disabled by default (it's only needed when the user wants to prepare examples with the same page layout features (for example to judge line breaking).
* *[Probably more to come ...]*

The double include issue could probably be solved by checking for such an include statement since we are in control over the filename.

A workaround from the other perspective would be to *not* require an explicit \include statement but teach Frescobaldi (creating an extension (using the new API)) to transparently inject it upon compilation. That would alleviate the "has to do it manually" issue but has substantial drawbacks: it would be application-specific, and it would essentially require a new "compilation mode" specifically for lyluatex.

@jperon What do you think?

jperon:
Have the headers written to a separate file so that users may include
them manually while writing on scores, and get the final result.


jperon:
@uliska Do you think this solves your issue ?

jperon:
@uliska Would you mind either merge or close this PR before #243 gets too far ? I'd like to avoid possible conflicts when changes get bigger.

uliska:
Sorry for leaving this so far behind - especially since it is about a request of mine.

I like the output in the file, and I think that part can be kept unchanged (although I didn't try very hard to test it with complicated option combinations).

However, I think it is somewhat fruitless to save this as to `output..'-headers.ly'` since that is not only inconvenient to access through `.list` file but in particular it is so volatile. The target filename will change with *any* modification to the LilyPond code and even the inclusion options.

I think including the headers really is only useful when preparing the score in an external file, and it will only be relevant in examples of such dimensions that they will typically be included as external files. So the headers should not be put in the `tmpdir` but in the user space. My suggestion would be one out of

* `<input-filename>-lyluatex-headers.ily`
* `path/to/directory/<filename>-lyluatex-headers.ily` (if a certain option is set)
* `path/to/explicit/filename.ily`

The second and third could be absolute or relative paths as usual (although one could argue whether relative paths should be relative to the LilyPond or the compiled LaTeX file).
The reasoning behind the last option is that in many cases when you have many examples in one document the headers will be identical. So what I would probably have in most projects is a fixed file as package option and override that locally when necessary (eg. when an example needs some indent or different staffsize).

---

Yes, this is a valuable thing, and I assume I'd use it all the time. In my current project it is not *that* much of a problem but in a number of cases the different sizes in Frescobaldi and LaTex just makes the difference between one-system and one-system-plus-one-measure scores, which would be great to be able to fine-tune in Frescobaldi.

Once this is in place here I will probably write some syntactic sugar for use in LilyPond (probably a tiny openLilyLib package) that can include the headers if *not* compiled through lyluatex.

*PS:* For testing I merged `master` into your branch, and that worked, so I suggest pushing that to minimize any risks of later merge conflicts

jperon:
@uliska Should be ready for merge: `write-headers` now takes a value, that's the destination dir of the file. For example, with this mwe, headers will be written to `headers/test-lyluatex-headers.ily`:
```tex
\documentclass{scrartcl}
\usepackage[write-headers=headers]{lyluatex}

\begin{document}
\lilypondfile{test.ly}
\end{document}
```
The path may be absolute (`/tmp/headers` would work); if not, it's relative to the *current* dir, which means the dir from which `lualatex` is launched (same as `tmpdir`).
Incidentally, this let me enhance `tmpdir`: this can now be absolute too.

Perhaps there would be conflicts with #243; never mind, merge it if it works, and then I'll deal with conflicts while reviewing #243.

uliska:
I'm out of town today, just wanted to let you know.
-- 

jperon:
@uliska If you agree with that, I think it should be ready for merge.

uliska:
I've just added a check to prevent an error when having a score without filename.

There's one more thing I'll add in a minute: an explicit filename for the headers, as detailed in https://github.com/jperon/lyluatex/pull/233#issuecomment-492398532

uliska:
@jperon Due to some more changes I'd like to give it back to you for final consideration and merging.

jperon:
@uliska The only changes I did are merely cosmetic, excepted that, instead of a global `lyluatex_opt`, I put a local `_opt` to minimize typing and avoid useless global variable.

uliska:
Well, I *did* intend to directly use that global variable in a depending follow-up package, but I see that it is cleaner to not do that and re-require the module a second time.

jperon:
IMHO, there should be two cases:
- either you want to get an option defined in `lyluatex.sty` or `lyluatex.lua`: then you'll get it from `Score`;
- or you want to use methods (`optlib.is_dim` and so on) or define a new set of options, and then yes, you should require the module again.

PR #236, opened 2019-05-01
Add system-count option 

uliska:
If given the `\paper { system-count }` option is set in the ly_code. This can be very useful when LilyPond would by itself break the music on too many systems (e.g. one full system plus one measure), where forcing the system-count can make it work.

Produces
```lilypond
\paper{
    system-count = 2
    two-sided = ##f
```
in the relevant portion of the generated ly file

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -1029,20 +1029,39 @@ Given: %s
 end
 
 function Score:ly_paper()
+    local function get_system_count()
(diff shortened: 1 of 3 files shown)


jperon:
I only added a slight optimization by avoiding (duplicate) function calls. Incidentally, I did the same for another local function.

Issue #237, opened 2019-05-09
Hash sign "illegal" in LilyPond


uliska:
```tex
\begin{lilypond}
#(ly:message "Hey")
\end{lilypond}
```
fails with

```
! You can't use `macro parameter character #' in vertical mode.
<recently read> ##
  
l.7 #
   (ly:message "Hey")
?
```
Obviously the `#` is interpreted as LaTeX (like we had it with e.g. `\repeat`) if the LilyPond snippet starts with it (if you place some regular LilyPond code before that it works like expected).

While this concrete example doesn't make sense it is perfectly reasonable to have snippets that only include one call to a music function which then produces the actual score, e.g. something like
```tex
\begin{lilypond}
#(example '(path to example))
\end{lilypond}
```

uliska:
Note that `\lilypond{#(example '(path to example))}` correctly passes the code to the LilyPond file.

jperon:
Are you sure you're using last version from the repo ? On my computer, this mwe compiles as expected : 

```latex
\documentclass{scrartcl}
\usepackage[debug]{lyluatex}

\begin{document}

\begin{lilypond}
#(ly:message "Hey")
\end{lilypond}

\end{document}
```

uliska:
Yes, I *am* using the latest pulled `master`. However, Probably the issue arised from having placed the code in a more complex context, and probably I was once more tricked by some residues in auxiliary files (I find this a pretty annoying "feature" in LaTeX).

Starting over with a fresh MWE and then with the content placed in an empty file with my project's documentclass I see that it *does* work properly. Sorry for the noise.

PR #239, opened 2019-05-09
Fix #238: problem when including empty scores.

### LINKED ISSUE #238: Illegal empty scores
Somehow related to #237 
LilyPond snippets that erroneously produce no score trigger an interesting side effect that we should handle gracefully.

```tex
\documentclass{article}
\usepackage{lyluatex}

\begin{document}

\lilypond{ #(ly:message "Foo") }

\end{document}
```
will fail with
```
lyluatex.
lua:1278: attempt to index a nil value
stack traceback:
	.../uliska/texmf/tex/luatex/git-latex/lyluatex/lyluatex.lua:1278: in function 
'write_latex'
	.../uliska/texmf/tex/luatex/git-latex/lyluatex/lyluatex.lua:1193: in function 
'process'
                                                  
l.7 \lilypond{ #(ly:message "Foo") }
                                  
?
```

The issue is that the LilyPond code produces no score. Of course this is erroneous input but we should handle it more gracefully with a *meaningful* message and not `attempt to index a nil value`.

The underlying issue is that by default LilyPond will not produce a PDF when there is no content:

```lilypond
\version "2.18.2"

#(display "Foo")
```

But

```lilypond
\version "2.18.2"
\include "lilypond-book-preamble.ly"
#(display "Foo")
```
*will* produce files - with the PDF being empty, without extent, and that seems to be where we get a `nil` value somewhere. *[NOTE: I currently have no LilyPond 2.18 installed so I'm not sure if that behaves identically.]*

I'm not sure about the best remedy. Probably there should be another check in `is_compiled()` or when checking the bounding box. From the program flow it should of course be when checking for the existence of the PDF, but maybe when checking the bounding box we have better information about the issue.

jperon:
What about this solution ?

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -1190,6 +1190,12 @@ function Score:process()
     else table.insert(self.output_names, self.output)
     end
     set_lyscore(self)
+    if self:count_systems() == 0 then
+        warn([[
+The score doesn't contain any music:
+this will probably cause bad output.]]
+        )
+    end
     if not self['raw-pdf'] then self:write_latex(do_compile) end
     self:write_to_filelist()
     if not self.debug then self:delete_intermediate_files() end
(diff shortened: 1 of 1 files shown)


uliska:
This looks fine, although I didn't test it.

OTOH, given the "thing" with #237 it may have been not strictly necessary, isn't it? Anyway, such a warning is definitely a good idea because in the majority of cases empty scores will be the result of erroneous input.

jperon:
Yes, I did have an error with `\lily` (as `self:bbox(1)` returned a nil value, and `self:bbox(1).height` was needed). So let's merge.

PR #240, opened 2019-05-09
Gracefully handle failed LilyPond start with 'showfailed' option

uliska:
I think the 'showfailed' option should not only cover cases when LilyPond
fails to produce a score but also when it can't be started.

With the showfailed option a failure to start LilyPond does not result
in an err() but in a warn() with the same text, and a box is written
into the document like the one reporting a failed score compilation.

With this option it is for example possible to have people successfully
compile a document without a proper LilyPond installation (useful when
they only care for the content or for technical assistance).

MWE:

```tex
\documentclass{article}
\usepackage{lyluatex}
\begin{document}
\lilypond[program=Foo,showfailed]{ c' }
\end{document}
```

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -935,15 +935,8 @@ function Score:lilypond_version(number)
                 self.output, self.program
             )
             debug(result)
+            return true
         end
-    else
-        err([[
-LilyPond could not be started.
-Please check that LuaLaTeX is started with the
---shell-escape option, and that 'program'
-points to a valid LilyPond executable.
-]]
-        )
     end
 end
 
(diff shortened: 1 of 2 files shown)


lyluatex.lua
```diff
@@ -1198,7 +1211,6 @@ end
 function Score:run_lilypond(ly_code)
     if self:is_compiled() then return end
     mkdirs(dirname(self.output))
-    self:lilypond_version()
```
uliska: I got slightly confused about this when reviewing my own staging area.

Am I right that this call was previously done exclusively to check for a successful LilyPond installation? If so it should be safe to remove it because that check is now done right at the entry of `Score:process()`.

@jperon

jperon: In any case, as no return value is used here, and `lilypond_version()` has already been run at this time, there's certainly no need to keep it here.

uliska:
The line had already been removed, with my comment I just wanted to draw specific attention to it.

PR #244, opened 2019-05-14
Document insert=system oddities

### LINKED ISSUE #193: Vertical spacing behaviour with include=systems
`include=systems` implies the use of `lilypond-book-preamble.ly`. It is worth pointing out that this will *not* have any notion of pages anymore - resulting in the staff-staff spacing to be minimal/natural. Usually LilyPond will space out the inter-staff (not -system!) space when the page is not filled, but with `include=systems` this will not happen.

While this behaviour is usually desirable when including score examples in text it may result in subpar output for multipage scores when there's the typical issues of how many systems will fit on a page.

Also it should be pointed out that by default pages will be ragged-bottom, which can be avoided by defining `\betweenLilyPondSystem` to somehow use `\vfill`.

These are not "bugs", but I think it should be part of the manual.

jperon:
Fix #193

### DIFF
--- a/lyluatex.md
+++ b/lyluatex.md
@@ -319,6 +319,19 @@ With this default option each score is compiled as a sequence of PDF files
 representing one system each. By default the systems are separated by a
 paragraph and a variable skip depending on the staffsize.
 
+\lyIssue{Note:}
+\option{insert=systems} implies the use of `lilypond-book-preamble.ly`.
+It is worth pointing out that this will *not* have any notion of pages
+anymore - resulting in the staff-staff spacing to be minimal/natural.
+Usually LilyPond will space out the inter-staff (not -system!) space
+when the page is not filled, but with \option{insert=systems}
+this will not happen.
+While this behavior is usually desirable when including score examples
+in text it may result in suboptimal output for multi-page scores when there's
+the typical issues of how many systems will fit on a page.
+Also notice that by default pages will be ragged-bottom,
+which can be avoided by defining \cmd{betweenLilyPondSystem} to somehow use `\vfill`.
+
 \lyCmd{betweenLilyPondSystem}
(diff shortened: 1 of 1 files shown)


lyluatex.md
```diff
@@ -319,6 +319,19 @@ With this default option each score is compiled as a sequence of PDF files
 representing one system each. By default the systems are separated by a
 paragraph and a variable skip depending on the staffsize.
 
+\lyIssue{Note:}
+\option{insert=systems} implies the use of `lilypond-book-preamble.ly`.
+It is worth pointing out that this will *not* have any notion of pages
```
uliska: Minor detail: I'd replace "this" with "the score" anymore to make it more clear.

I suggest "pages and page breaking anymore"

lyluatex.md
```diff
@@ -319,6 +319,19 @@ With this default option each score is compiled as a sequence of PDF files
...
+anymore - resulting in the staff-staff spacing to be minimal/natural.
+Usually LilyPond will space out the inter-staff (not -system!) space
+when the page is not filled, but with \option{insert=systems}
+this will not happen.
+While this behavior is usually desirable when including score examples
+in text it may result in suboptimal output for multi-page scores when there's
+the typical issues of how many systems will fit on a page.
+Also notice that by default pages will be ragged-bottom,
```
uliska: Suggestion: insert something like “and LilyPond will not make any efforts to optimize page breaks.“

lyluatex.md
```diff
@@ -319,6 +319,19 @@ With this default option each score is compiled as a sequence of PDF files
...
+which can be avoided by defining \cmd{betweenLilyPondSystem} to somehow use `\vfill`.
```
uliska: ... and in addition to the comment above: “\cmd{betweenLilyPondSystem} can also be used when the space between systems seems to be too tight”.

PR #247, opened 2019-05-18
optlib.merge_options

uliska:
A helper function to create a table with current options superseded
by local options. This is not needed in lyluatex where the same result
is achieved through the Score object, but typically a client is required
to perform this merge for *any* score (or other element).

@jperon there may be a few more minor PRs coming, although I really have to get back to writing *content*. But I want to drive my [lyluatexmp](https://github.com/uliska/lyluatexmp) package to the initial point where I can replace the old solution with the new package (i.e. have one command `\lyfilemusicexample` ready to be used). In that process I find that a number of tool functions “appear“ that need to be factored out, some to my own helper modules but some also to `lyluatex`. Maybe it is a good idea to merge these modules to one common LaTeX package at some point (when `lyluatex` has been through its next proper release and my own project is submitted).

### DIFF
--- a/lyluatex-options.lua
+++ b/lyluatex-options.lua
@@ -255,7 +255,7 @@ function optlib.merge_options(base_opt, super_opt)
     super_opt. Entries in super_opt supersede (i.e. overwrite)
     colliding entries in base_opt.
 --]]
-    result = {}
+    local result = {}
     for k, v in pairs(base_opt) do result[k] = v end
     for k, v in pairs(super_opt) do result[k] = v end
     return result
--- a/lyluatex-options.lua
+++ b/lyluatex-options.lua
@@ -248,5 +248,19 @@ function optlib.is_str(_, v)
 end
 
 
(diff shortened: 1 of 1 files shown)


jperon:
@uliska "If I were you", I think I'd rather use the metatable and `__index` mechanism of Lua; that said, this function certainly may have use cases, so no objection for merge.

uliska:
I think there are numerous valid use cases to write a client using `lyluatex-options.lua` whose needs are limited and where the extra effort of making the client object oriented is not really necessary.

In “my” case this is probably different, yes. I assume I'll stick to LaTeX for quite some time and will increasingly work on the Lua side of things because that's just so powerful and “clean“ (compared to TeX programming), especially when thinking about *generation* of things like composing a document from database entries or the like. One major task I'm still going to need is a package to typeset critical reports (from data exported from LilyPond). So it should really be worth the effort getting my head around how Lua is designed, and how it interacts with LaTeX.

PR #248, opened 2019-05-20
Let check_local_options handle nested curly braces

uliska:
Nested curly braces in an optional argument break the existing code, if
a closing brace happens to be followed by a comma, like in
```tex
caption={Score with \emph{two}, three or four bars}
```
This commit keeps track of the nesting level to ensure the whole argument
is stored as a single option value.
(Faulty input will result in follow-up errors, but these should be clear
from the LaTeX error messages).

### DIFF
--- a/lyluatex-options.lua
+++ b/lyluatex-options.lua
@@ -70,13 +70,34 @@ function Opts:check_local_options(opts, ignore_declarations)
     step is skipped (i.e. local options are only parsed and duplicates
     rejected).
 --]]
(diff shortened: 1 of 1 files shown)


jperon:
@uliska Could you please check it works with last commit ? If it does, please "squash and merge" it.

uliska:
Yes it does. Thank you for the concise solution. I *did* in fact doubt my solution was the best implementation.

PR #249, opened 2019-05-22
Retry lilypond command when failed

### LINKED ISSUE #204: LilyPond cannot be found
Hello,

I heard of this package in a seperate issue for TeXstudio and it seems to be exactly what I am looking for. Problem is, TeXstudio can't seem to find LilyPond in order to use it.

[https://github.com/texstudio-org/texstudio/issues/236](url)

I added LilyPond to the environmental path, as shown here [http://lilypond.org/windows.html](url). Unfortunately, the error persists. Can the package be tailored for Windows users, or at least some form of hint given, how Windows users can use this package?

```
Der Befehl "lilypond" ist entweder falsch geschrieben oder konnte nicht gefunden werden.

...ta/Local/Programs/MiKTeX 2.9/tex/latex/base/ltluatex.lua:109: 
Module lyluatex Error: LilyPond could not be started.
(lyluatex)             Please check that LuaLaTeX is started with the
(lyluatex)             --shell-escape option, and that 'program'
(lyluatex)             points to a valid LilyPond executable.
(lyluatex)             on input line 35
```

jperon:
Fix #204: MiKTeX behavior isn't the same as TexLive's with io.popen. I
don't know why, but on MiKTeX, we have to add a level of quotes, and
it doesn't seem to be able to write code to stdin.

Incidentally, when LilyPond fails on a score, compilation is re-tried
with debug=true.

### DIFF
--- a/lyluatex-lib.lua
+++ b/lyluatex-lib.lua
@@ -135,4 +135,29 @@ function lib.splitext(str, ext)
         or (str:match('(.*)%.(%w*)$') or str)
 end
 
+
+------------------------------------
+-- Engine, version, TeX distribution
+
+local tex_engine = {}
+setmetatable(tex_engine, tex_engine)
+
+function tex_engine:__call()
+--[[
+    Defines the properties extracted from the first line of jobname.log.
+--]]
+    local f = io.open(tex.jobname..'.log')
+    if not f then return end
+    self.engine, self.engine_version, self.dist, self.dist_version, self.format, self.format_version =
+        f:read():match(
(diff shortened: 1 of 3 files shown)


jperon:
A drawback of this is that, in case of LilyPond failures, there will be misleading messages about unrecognized commands. IMHO, it isn't such an issue.

jperon:
@uliska When it's merged, I think we could release 1.0 final.

uliska:
@jperon sorry I didn't process anything the last days.

For a final release we should go through the manual once more to check if anything was missed or to find any remaining flaws.
-- 

jperon:
@uliska Do you think you could do that ? I must confess that doc isn't my cup of tea…

uliska:
I'll have a look and see what happens...
-- 

PR #250, opened 2019-05-26
Manual: Review introductory texts

uliska:
I don't have the time to fully review the manual, which by now contains lots of text. What I *did* is review the introductory texts, hopefully giving a smoother introduction to people and giving those a better chance of understanding why it should be of interest who don't know yet what to make of that package.

With regard to the remaining text I think we can assume everything is OK. What *might* be a good idea, though, is going through the list of options in `lyluatex.sty` and make sure each of them is covered in the manual.

### DIFF
--- a/lyluatex.md
+++ b/lyluatex.md
@@ -2,6 +2,7 @@
 documentclass: lyluatexmanual
 title: "\\lyluatex"
 subtitle: "1.0b"
+date: \today
 author:
 - Fr. Jacques Peron
 - Urs Liska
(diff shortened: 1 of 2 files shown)


jperon:
@uliska I just added a commit to change how the date of the manual is calculated: instead of current date at time of compilation, it fetches the date of last commit.

jperon:
@uliska I just pushed documentation of undocumented options.

uliska:
>  instead of current date at time of compilation, it fetches the date of last commit.

That's a nice one :-)

PR #251, opened 2019-05-28
Move support modules to lyluatextools repository

uliska:
@jperon this is a sketch for how I thought we could factor out the modules  to a separate support repository. In order to work you have to clone the [lyluatextools](https://github.com/uliska/lyluatextools) repository

The support modules are moved to the new lyluatextools repository.
Requiring lyluatexoptions will provide the lyluatex_options global variable.

Requiring lylautexlib will provide the lyluatex_lib global variable
(not used in lyluatex).

Requiring lyluatextools would provide both variables.


uliska:
PS: I'd be totally fine with
*  moving ownership of `lyluatextools` to hte `jperon` account
* ditching the separate repository but still create `lyluatexoptions.sty` and `lyluatextools.sty` (in the `lyluatex` repo

jperon:
I’m not opposed to the *principle* of a new repository for that, but I think that for now, integrating `lyluatextools.sty` (and possibly `lyluatexlib.sty` and `lyluatexoptions.sty`) within `lyluatex`’s one would be a better option for now, as:

- a new repository would mean a new dependency for `lyluatex`, so a little bit more complicated for users who would like to install it manually;
- for now, only `lyluatex` and `lyluatexmp` (that depends on `lyluatex`) use it; I’d understand a separation if a package not depending on `lyluatex` used it.

As for the packages, I’m wondering whether having 3 ones wouldn’t be overkill: why not, instead, make one package, having an option to decide whether `lyluatex_lib`, `lyluatex_options` or both are defined ?

As for the ownership, it's been a while since I've been thinking about transferring `lyluatex` to [openlilylib](https://github.com/openlilylib/), since it has grown up far beyond my personal use. Another option would be to create a special organization for it (as [gregorio-project](https://github.com/gregorio-project) for `Gregorio`). What do you think of it ?

jperon:
@uliska What about implementation in #252 ?

PR #252, opened 2019-05-29
lyluatextools

jperon:
All options in one auxiliary package.
Closes #251.

### DIFF
--- a/lyluatex.sty
+++ b/lyluatex.sty
@@ -17,6 +17,7 @@
 \RequirePackage{currfile}
 \RequirePackage{pdfpages}
 \IfFileExists{varwidth.sty}{\RequirePackage{varwidth}}{}
+\RequirePackage[options]{lyluatextools}
 
 \RequirePackage{metalogo}
 \newcommand{\lyluatex}{\textit{ly}\LuaTeX}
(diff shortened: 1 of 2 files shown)


uliska:
@jperon I think on the long run we should talk about moving that code to a completely independent package (also without the `lyluatex` part in the name). At least I will immediately make use of it in packages that *don't* make use of `lyluatex`. And I think this could become a generally useful toolkit for developing LuaTeX packages. I don't have sufficient experience with that but it might be worth raising this on the luatex mailing list, what do you think?

jperon:
@uliska I think it's a good idea to raise it on the luatex mailing list. I see at least two points, especially with `lyluatex-lib.lua`:

- how to deal with versions ? Some breaking changes could become a real headache if some packages depend on it (that's a recurring problem with libraries in general);
- some "`stdlib`s" for lua already exist: [penlight](http://stevedonovan.github.io/Penlight/api/), [lua-stdlib](https://github.com/lua-stdlib/lua-stdlib)... Wouldn't it be worth creating a package for one of those on CTAN, rather than inventing a new one ?

jperon:
@uliska I've just checked, and seen that there already is a `lualibs` package on CTAN. I didn't look further, but I'm fairly sure it would cover most of what `lyluatex-lib` does.

uliska:
OK, one day I'll look into that @jperon . Seems like we should better see whether we can *use* such existing functionality wherever we can avoid writing custom functions.

PR #261, opened 2019-07-15
\lyuseoption as wrapper to Opts:use_option

uliska:
I've come to the conclusion that it's a courtesy to the user to
reduce the need to use \directlua wherever possible ...

Use case: In a document preamble it is with this patch possible to write

```tex
\RequirePackage[%
    left=\lyuseoption{my_opts}{layout_left},%
    right=\lyuseoption{my_opts}{layout_right},%
]{geometry}

rather than having to call `my_opts:use_option(layout_left)` in a `\directlua` clause
```

### DIFF
--- a/lyluatex-options.lua
+++ b/lyluatex-options.lua
@@ -158,6 +158,12 @@ function Opts:use_option(key)
     end
 end
 
+-- Wrapper macro to simplify the use of Opts:use_option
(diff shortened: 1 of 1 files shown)


uliska:
I will also add this to `luaoptions` but only close this PR when I've done so.

PR #262, opened 2019-07-18
Allow non-package option sets

uliska:
With this commit it is possible to add an option set
without making it package options.
The use case is in luatemplates to set up an options
object for individual formatters to validate optional
arguments in macros that are not package options but
specific to the macro.


uliska:
This has become obsolete as a PR against `lyluatex` and is now part of `luaoptions`

PR #264, opened 2019-07-22
Move lyluatextools dependency to new package luaoptions

uliska:
I think it is definitely time to pull the option handling code out into a separate package.
I've been working on a new package [`luaformatters`](https://github.com/lualatex-tools/luaformatters), and this confirms what I had already thought for some time: There are lots of uses for the option handling functionality that don't justify *lyluatex* as a dependency. Having `luaoptions` as a standalone package (and repository) will also likely increase its visibility and use and through this maybe support and additional contributors. [`luaoptions`](https://github.com/lualatex-tools/luaoptions) has to be cloned from https://github.com/lualatex-tools/luaoptions.git or git@github.com:lualatex-tools/luaoptions.git to test this Pull Request.

Right now `luaoptions` is basically a fork stripped off the lyluatex-related files, but I suggest moving the generic repositories/packages to a new organization.

For now the new package has barely any modifications because I want to have an agreed-upon transition before starting to think about further modifications. The only thing I have silently done during the extraction is integrate the commit from https://github.com/jperon/lyluatex/pull/262 because that's crucial for me to work with (and I hope it's actually non-controversial).


uliska:
It is clear that the check build fails because I've removed essential files from the repository and this isn't reflected in the test set-up.

jperon:
@uliska Would you mind looking at `.travis.yml` ? I think there are commands to add into the before_install section, but I won't have time to look at it for several days. If you can't get to fix it, never mind and merge ; I'll look at it afterwise. The same goes for other pull requests : as I understand it would ease your work to merge them quickly enough, don't wait for me to look at them ; I'll submit corrections later if necessary.

uliska:
> @uliska Would you mind looking at .travis.yml ? I think there are commands to add into the before_install section, but I won't have time to look at it for several days.

I'll do so ASAP (but today I'm out of town). Basically what has to be done is clone the `luaoptions` repository to a place where `lualatex` can find it.

> The same goes for other pull requests : as I understand it would ease your work to merge them quickly enough, don't wait for me to look at them ; I'll submit corrections later if necessary.

Thank you, I'll take that liberty with the luaoptions package. For now I don't need the Pull requests merged in `lyluatex` anymore. I suggest that we keep this current PR open for a while until the interface in luaoptions has settled. So if any changes might also require the modification of clients *using* `lyluatex` that would happen in one atomic step.

uliska:
OK, it builds correctly now, yippieh.
But still I suggest keeping the PR open until it has become clear that no further changes to the interface are necessary.

uliska:
I have pushed a few commits adapting the code to changes in the `luaoptions` modules. The commit messages include links to the corresponding commits in `luaoptions` (FYI and reference, @jperon).

It seems that when having the option handling package as a standalone tool some things are appearing in a different light, and it seems to be right to do a few things slightly differently (for example: storing the `Opts` object not in global variables (polluting the Lua namespace) but in one local table and provide an accessor function ...)

PR #292, opened 2022-08-05
Fix count_systems

### LINKED ISSUE #290: Missing final system on recompile
I have found that on the `master` branch of lyluatex, I often get an erroneous output if I compile my document twice. For a trivial example, I have two files in a directory:

test.tex:
``` tex
\documentclass{minimal}

\usepackage{lyluatex}

\begin{document}
    \lilypondfile{test.ly}
\end{document}
```

test.ly:
``` lilypond
melody = \relative c'' {
  g1 | \break
  g1 | \break
  g1 |
}

\score {
  \new Staff  <<
    \melody
  >>
}
```

The first time I compile it I get this result with all three systems:

image

But if I compile it again (without changing anything) I only get two systems:

image

No further recompilations give me any different results. I am stuck with only two systems. I have to remove the `tmp-ly` directory to get all three systems again.

I found that this bug does not seem to exist in the 1.0 version, so I used git bisect to try and find the guilty commit. It only seems to occur in the most recent commit (06a7c270ec2251d4ad6bcda016879689dd4e6484).

I have not had any time to delve in any deeper, or try to root cause it yet. This is my first time using lyluatex (I am new to lilypond in general), so I wasn't positive that this wasn't user error, but it seems pretty clearly erroneous to me.

I am running Ubuntu 20.04.2 LTS under WSL2 and using it's version of tex-live.

Cheers 🍻

jperon:
fix #290

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -702,18 +702,18 @@ function Score:content()
 end
 
 function Score:count_systems(force)
-    local count = 0
+    local count = self.system_count
     if force or not count then
-        local systems = self.output:match("[^/]*$").."%-?%d*%.eps"
+        count = 0
+        local systems = self.output:match("[^/]*$").."%-%d+%.eps"
         for f in lfs.dir(self.tmpdir) do
             if f:match(systems) then
                 count = count + 1
             end
         end
-        if count > 1 then count = count - 1 end
         self.system_count = count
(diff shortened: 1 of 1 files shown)


sudobash1:
That fix works well for me. I appreciate it!

PR #296, opened 2022-10-30
CTAN release

jperon:


jperon:
[CTAN update accepted](https://www.ctan.org/ctan-ann/id/mailman.1667.1667243826.3715.ctan-ann@ctan.org)

PR #298, opened 2022-11-04
Fix compatibility with polyglossia

### LINKED ISSUE #297: \setmainfont causes problems
After adding to a quite complicated preamble lyLuaTeX I received a bunch of errors. They are caused by  \setmainfont{Minion Pro}. Scores are compiling correctly but the font is set to standard. The file compiles correctly with no lyLuaTeX and with no \setmainfont. But when both are in the file I have such error message:

```
...al/texlive/2022/texmf-dist/scripts/lyluatex/lyluatex.lua:1350: attempt to in
dex a nil value (field 'rawdata')
stack traceback:
	...al/texlive/2022/texmf-dist/scripts/lyluatex/lyluatex.lua:1350: in function 
'lyluatex.lua.get_font_family'
<argument> \directlua {ly.get_font_family(font.current())}
                                                
l.94 \end{lilypond}
                 
? 
...al/texlive/2022/texmf-dist/scripts/lyluatex/lyluatex.lua:1350: attempt to in
dex a nil value (field 'rawdata')
stack traceback:
	...al/texlive/2022/texmf-dist/scripts/lyluatex/lyluatex.lua:1350: in function 
'lyluatex.lua.get_font_family'
	...al/texlive/2022/texmf-dist/scripts/lyluatex/lyluatex.lua:1364: in function 
'lyluatex.lua.set_fonts'
\ly@currentfonts ..., \sffamilyid , \ttfamilyid )}
                                                  \endgroup 
l.94 \end{lilypond}
                 
? 
```

jperon:
Fix #297

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -1347,7 +1347,7 @@ end
 
 
 function ly.get_font_family(font_id)
-    return lib.fontinfo(font_id).shared.rawdata.metadata['familyname']
+    return lib.fontinfo(font_id).fullname
 end
 
 
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -1347,7 +1347,7 @@ end
 
 
 function ly.get_font_family(font_id)
-    return lib.fontinfo(font_id).fullname
+    return lib.fontinfo(font_id).fullname:match("[^-]*")
 end
 
 


tomaszgrabowskiop:
It gives the most interesting result... No errors (great!). But it changes a font to some I don't use for sure... Look into attached fragments. First as it was with Minion and second what shows after a change you made. 

[8bc13bab4bc1db70258b581f107eb559-1.pdf](https://github.com/jperon/lyluatex/files/9944979/8bc13bab4bc1db70258b581f107eb559-1.pdf)
[5593b8589b544f4bc131f2fa57c3c8d3-1.pdf](https://github.com/jperon/lyluatex/files/9944980/5593b8589b544f4bc131f2fa57c3c8d3-1.pdf)

jperon:
May you please add this to your MWE:
```tex
\directlua{tex.sprint(font.getfont(font.current()).fullname)}
```

tomaszgrabowskiop:
It only has added "Minion Pro-Regular" but the font in lyrics is at it was.

jperon:
What’s weird here is that it seems to mean lyluatex passes the good font name, but LilyPond doesn’t find it. Are the files of Minion Pro installed on your system (for example, in Debian / Ubuntu, in `/usr/share/fonts/` or `$HOME/.local/share/fonts/`, or only in the folder of your document ?

tomaszgrabowskiop:
We are talking about MacOSX 12 in my case. The font is installed for user. 
`/Users/XXX/Library/Fonts/MinionPro-Regular.otf`

Before the change you did in lyluatex.lua the font was visible for fontspec package. But it was necessary to invoke it before and besides polyglossia. Now in all cases lily pond sets the lyrics with font I don't use in tex document.

jperon:
Could you please:

- remove the `tmp-ly/` directory;
- compile your document;
- zip the whole folder and [send it to me](mailto:cataclop@hotmail.com?subject=Polyglossia%20bug%20(#298)) ?

jperon:
@tomaszgrabowskiop I’ve just pushed a commit (on this branch) that should IMO fix the problem. The error came from the difference between `fullname` (e. g. Minion Pro-Regular) and `family name` (e. g. Minion Pro). The font with which I work (`libertine`) returns the same value for both, so I hadn’t noted this difference!

tomaszgrabowskiop:
It does the trick. Thank you!

tomaszgrabowskiop:
Even when polyglossia is turned on and fontspec is not load separately.

jperon:
I think this point is on LilyPond side: `lyluatex` only defines fonts thank to [make-pango-font-tree](https://lilypond.org/doc/v2.23/Documentation/notation/fonts.html), LilyPond (and not LaTeX) formats the lyrics.

lfm83:
This commit (up to current master) breaks pass-fonts for me.
If I use
`\setmainfont{EB Garamond}`
then in the created temporary .ly file, we get a font tree looking like:
`(make-pango-font-tree "EB Garamond Regular" [etc]`
the "Regular" part of which should not be there: LilyPond finds the EB Garamond only without the "Regular" specification.

jperon:
@lfm83 OK, I see the problem. I’m going to fix it, but there will be a drawback: it will only work with `EB Garamond` without `polyglossia`, or at least only by loading `polyglossia` after using `setmainfont`. The reason is that `polyglossia` removes the `rawdata` table from the font table from `luatex`, thus making it impossible to properly get the `family name` of the current font.

jperon:
@lfm83 Should be fixed in latest master.

lfm83:
> @lfm83 Should be fixed in latest master.

Yes, I can confirm the bug is fixed for me. Thank you!

PR #300, opened 2022-11-20
support MiKTeX with LilyPond ≤ 2.22

szollbrecht:
These changes made it work for me on windows 11 with miktex, running lualatex from the console or texniccenter.

It's more a workaround than a proper solution but maybe it helps you finding the issue with miktex or someone else in the meantime.

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -105,7 +105,13 @@ end
 
 local function extract_includepaths(includepaths)
     includepaths = includepaths:explode(',')
-    local cfd = Score.currfiledir:gsub('^$', './')
+
+    if lib.tex_engine.dist == 'MiKTeX' then
+        local cfd = Score.currfiledir:gsub('^$', '.\\')
+    else
+        local cfd = Score.currfiledir:gsub('^$', './')
+    end
+    
     table.insert(includepaths, 1, cfd)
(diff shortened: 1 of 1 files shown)


jperon:
I intended to try it yesterday, so I launched a Windows 10 VM, and tried to install MiKTeX. But when I see how it works on my first MWE and [how the main author handles issues](https://github.com/MiKTeX/miktex/issues/1249#issuecomment-1357247758), I suddenly feel less motivated…

jperon:
After eventually finding a way to have MiKTeX run (my Windows 10 had to be updated), I have checked that this fix is working *with lilypond 2.22*. With 2.24, there is another problem (seems a problem with pdf version) that I can’t reproduce on TeXLive/Linux. As I said, having experimented the way issues are handled from MiKTeX, I feel no motivation to tell about this problem upstream; feel free if you want to.

So I’ll merge it as is, and add a documentation note about the incompatibility with 2.24.

szollbrecht:
oh my, really not helpful at all... "works for me"🤦‍♂️
thanks anyway, maybe in the future when I have motivation again, I will check once more :)

jperon:
@szollbrecht Thank you very much for having made the work!

PR #321, opened 2024-07-13
Add tall-page-formats option

### LINKED ISSUE #307: Some weird alignment issues when using a custom header
I've got some weird alignment issues when using a custom header.  The below project shows the problem on the second score (where the header field that usually appears on the left has been commented out).

[test.zip](https://github.com/jperon/lyluatex/files/10972997/test.zip)

This is new with LilyPond 2.24.  The code worked fine before upgrading my LilyPond (and lyluatex) versions.

Version information is as follows:
LyLuaTeX: 1.1.3
LuaTex: Version 1.15.0 (TeX Live 2022)
OS: macOS 12.6.3
Lilypond: 2.24.1 (running Guile 2.2)

rpspringuel:
LilyPond 2.24 introduced a change to the way the eps and pdf files are produced.  In particular, when using the lilypond-book preamble (a version of which is used in this package) to produce separate files for each system, the pdf file that included all the music contained one page per system while the eps file that include all music was a single page.  This meant that if the first line of the score was significantly different from the subsequent lines (e.g., a header with material flushed right instead of a music stave), then scores would get placed in unexpected positions.  This was the result of the eps bounding box and the pdf bounding box being based on different parts of the music (the whole score in the former case, and only the first line in the latter).

By adding the `-dtall-page-formats=pdf` option to the LilyPond compilation, the pdf containing all the music will put all the music on a single page, thus matching the eps file and allowing the bounding box calculations to be use more comparable values.

Fixes #307

### DIFF
--- a/lyluatex.lua
+++ b/lyluatex.lua
@@ -808,7 +808,7 @@ function Score:lilypond_cmd()
     end
     local cmd = '"'..self.program..'" '
         .. (self.insert == "fullpage" and "" or "-E ")
-        .. "-dno-point-and-click -djob-count=2 -dno-delete-intermediate-files "
+        .. "-dno-point-and-click -djob-count=2 -dno-delete-intermediate-files -dtall-page-formats=pdf "
     if self['optimize-pdf'] and self:lilypond_has_TeXGS() then cmd = cmd.."-O TeX-GS -dgs-never-embed-fonts " end
     if self.input_file then
         cmd = cmd..'-I "'..lib.dirname(self.input_file):gsub('^%./', lfs.currentdir()..'/')..'" '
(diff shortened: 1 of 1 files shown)


jperon:
@rpspringuel Thank you very much for this PR. May you please test with the small change I just pushed, which restricts the change to LilyPond >= 2.24?

rpspringuel:
I don't have an earlier version of LilyPond installed to check that behavior, but with 2.24.3 your change doesn't stop the fix from working for me.

jperon:
Thank you very much again!

lukas-vlcek:
Is this fix part of any release?

As of writing I can see that the official [latest release](https://github.com/jperon/lyluatex/releases/latest) points to https://github.com/jperon/lyluatex/releases/tag/v1.1.3 which has been released in Mar 1st, 2023. Does that mean that this fix is still not available to users?

If that is the case what can be done to make this fix available to users? Or is there any hotfix/workaround that can enable me to use this code? (Does it mean I would need to compile the code locally and manually install into my local TeX Live installation?)

Anyway, thanks for this marvellous package!

PR #326, opened 2025-03-14
Fix extra left padding for inline scores

derekerdmann:
Several lines that use `\setluaoption` did not end with a comment, introducing extra left padding to inline Lilypond scores.

With this related fix in the `luaoptions` package, no additional left padding gets inserted for inline scores with zero `hpadding`: https://github.com/lualatex-tools/luaoptions/pull/8


jperon:
Thank you!

